PR #16, opened 2019-04-30
Cleanup client if fail to Connect

alsm:


ovaltzer:
quick question regarding this merge - what is the outcome of the version without the cleanup? I mean into what situations can this take the client into?

alsm:
Prior to this change it was not possible to call Connect again on a client that had been disconnected or failed to connect properly. the cleanup function specifically ensures that we shouldn't leak goroutines with the ping service still running or the incoming() goroutine.

Issue #26, opened 2020-07-23
Support automated reconnection (and other actions when the network connection drops)


MattBrittan:
**Is your feature request related to a problem? Please describe.**

When the network link with the MQTT broker drops there is currently no obvious way to detect or handle this. There is an `OnDisconnect` call back but this is only called when the broker cleanly disconnects (i.e. a Disconnect notification is received from the broker). I believe that the most common reason to want to detect the loss of connection is so the connection can be re-established so that is the use-case I am going to focus on (but any solution must provide a means of detecting loss of connection).

The lack of an "obvious way" for a user to implement autoreconnection functionality with the library as it currently stands does not mean that it's impossible; but it would be difficult. The challanges I see are:

* When Publish returns an error how do I tell if the message has been stored and will be retried when the connection returns (as the store is not yet implemented this is a bit academic).
* It would be difficult to determine if any errors returned are connection related or due to something else (you could ignore those and just use the keepalive but I don't think that is a good approach)
* If the connection has gone down and we want to reconnect then we will need a threadsafe way of doing so. It is easy to introduce bugs here because there may be many goroutines attempting to call `Publish`, the store needs to be cleanly shutdown/reopened and we need to bring up a new connection.
* The connection may be down for some time and we may want to use the inbuilt store rather than implementing our own form of persistence (I would also like to be able to call `Publish()` before `Connect()` and have the message added to the store).

Because automatic reconnection is a common requirement I don't think its appropriate to leave all of the above to the user. This project should provide a method and/or recommended approach.

Note: Some of how this works will depend upon how persistence is implemented. The MQTT [spec dictates](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901238) that there are situations where unacknowledged PUBLISH (and related messages) MUST be resent. How this resend is handled will have an impact on how reconnections can be processed.

*Apologies - I may have gone overboard on the detail in this issue! However I believe its important to get this right because it has a significant impact on the ease-of-use of the package*

**Describe the solution you'd like**

My proposed solution contains three elements:

1. A clear way of telling if an error can be safely ignored (because the message will be delivered when the connection comes up).
2. A means of detecting that the connection has been lost.
3. A clean way of re-establishing a connection (making some assumptions about the store).

***Error Handling***

Any errors returned by a function that has stored the relevant message (e.g. in `MemoryPersistence`) will be wrapped by a `MessagePersistedError` i.e.: 

```
[...]

Issue #25, opened 2020-07-22
Persistence is not implemented


MattBrittan:
**Describe the bug**
The package includes a `Persistence` interface along with two implementations (`noopPersistence ` and `MemoryPersistence`) but these are not currently used. I have classed this as a bug rather than a feature request because it's partially implemented (and therefore likely to be confusing to new users).

**To reproduce**
Attempt to use the Persistence implementations.

**Debug output**
N/A

**Expected behaviour**
Messages persist as per the MQTT spec.

**Software used:**
- Server was - N/A
- Client version 7369e71

**Additional context**
Mentioned in #23

MattBrittan:
I am taking another look at the implementation of persistence. A few initial thoughts:
1. Initially only outgoing persistence will be implemented (on the basis that the PR enabling user control of ACK's will be implemented so it will be possible to effectively use the broker to handle inbound persistence). This will mean that the package will not fully comply with the spec (and QOS 2 will guarantee no duplication) but keeps things simple (it will be possible to add this later).
2.  Persistence handling will be completely user definable. My initial thought is that all inbound and outbound comms will be added to a channel (rather than being handled/transmitted directly) and the routine that processes these messages could be replaced (which would provide maximum flexibility, including the functionality requested in #40).

Anyway thought I should add a note to indicate that there are plans to do this (no guarantee on a time-frame sorry) ; if anyone has any thoughts/requirements then please let me know so I can feed these into the design.

yanmxa:
@MattBrittan Do you have a plan for how long to enable the `persistence` or `cleansession=false` in this repos?

We've been trying to use this session feature recently with this library, and I've also integrated this MQTT protocol into [CloudEvents](https://github.com/cloudevents/sdk-go), maybe more people will pay attention to this later.

MattBrittan:
@yanmxa I'm actually working on it right now. However I don't know if my current attempt will be successful (or, if it is, how long it will take). The challenge is implementing this in a way that keeps the library simple and idiomatic (really want to avoid the complexity of the v3.1 client).

ashtonian:
is there an equivalent clientconfig.SetConnectionLostHandler() ?

MattBrittan:
> SetConnectionLostHandler

Your comment does not seem directly related to this issue? Paho has [`OnClientError`](https://github.com/eclipse/paho.golang/blob/master/paho/client.go#L51) which will be called if there is a network error.

PR #27, opened 2020-07-23
Wiring up Persistence and looking at conn loss

alsm:
I've removed the OnDisconnect callback and moved to OnConnectionLost
instead and used a specific Error type to indicate when it was a server
initiated disconnect (see test change)

#25 
#26


alsm:
yeah, it's not done yet, thanks for the review I'll take a look

Issue #29, opened 2020-08-20
Automatic topic alias handling


Tockra:
**Is your feature request related to a problem? Please describe.**
At the moment you need to define the topic alias for each topic in every Publish:
```go
		// Computes the topic alias of the current topic (mqttv5 feature)
		if topicMap[topic] == 0 && lastTopicAlias < 65535 {
			topicMap[topic] = lastTopicAlias + 1
			lastTopicAlias += 1
		}

		topicId := topicMap[topic]
		var topicProp *mqtt.PublishProperties
		if topicId != 0 {
			topicProp = &mqtt.PublishProperties{TopicAlias: &topicId}
		}

		content := mqtt.Publish{
			// message reachs at least one time the target
			QoS: 1,

			// Broker doesn't safe the last message for new connections.
			Retain: false,

			// topic
			Topic: topic,

			// Properties (TopicAlias,...)
			Properties: topicProp,

			// Content
			Payload: msg,
		}
```

I think this is the common case that you wish to add a mapping between topic and a random id. 

**Describe alternatives you've considered**
Why isn't that implemented by default? If you active autoTopicAlias=true your MQTT client will give your topic the next free topic id?

Tockra:
By the way: Is the implementation of topic aliases in this library correct?
In https://github.com/eclipse/paho.golang/blob/7369e711591a4861f6091c1b0f48e8dbdb091e0a/packets/packets.go#L244 the Topic will added to the message. If you set the topic alias it will set to. But shouldn't it be removed after the first send with topicString and topicAlias ?

alsm:
As the topicaliases are not automatically handled for client publishes then the library implementation is correct, it is on the user to not send future publishes with the topic string set (unless they are looking to reregister the topic alias). Given the reregistration I'll have to think if there is a way to handle this. Generally though this implementation, unlike the v3 client, intends to expose more of the internals of the protocol to the library user and let them make more choices on how to handle such issues.

I'll certainly have a think about if there's a way to offer something like this without too much magic though, and I probably haven't handled the receiving of aliases from the broker properly.

alsm:
@Tockra Have a look at the above and let me know what you think. I've added a hook to the clientconfig that if set is called before publishes are sent and an extension that uses this to do automatic topic aliasing.

alsm:
Have implemented this, closing

PR #31, opened 2020-10-06
Add extension to provide automatic topic aliasing

alsm:
Added a publishHook function to clientconfig that if set is called
before the publish is sent, this function can modify the Publish and is
used in the extension to do automatic topic aliasing

#29


alsm:
No comments and it looks good to me so merging

Issue #32, opened 2020-12-13
The session expiry interval should be valid for CONNACK.


DrmagicE:
Currently, I am using this lib to test my MQTT [broker](https://github.com/DrmagicE/gmqtt). So I found this bug. I will make a PR to fix it. 
https://github.com/eclipse/paho.golang/blob/9638458a263ed04d87c32631ecfa5f33485d7938/packets/properties.go#L677

reference: https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901082

alsm:
Thanks for the PR, you need to sign the ECA to make PRs that I can merge. Also for the CONNACK there are a couple of other places it's referenced so I've added those to my own change. Good luck with your broker.

PR #33, opened 2020-12-13
fix: session expiry interval is valid for CONNACK

DrmagicE:
This PR will fix https://github.com/eclipse/paho.golang/issues/32


alsm:
Thanks for the PR, you need to sign the ECA to make PRs that I can merge. Also for the CONNACK there are a couple of other places it's referenced so I've added those to my own change. Good luck with your broker.

Issue #40, opened 2020-12-25
how about adding hook for PUBXXX messages


ryskiwt:
Dear maintainers,

**Is your feature request related to a problem? Please describe.**
I want to set user properties on PUBACK, PUBREC, PUBREL and PUBCOMP.

**Describe the solution you'd like**
How about adding a hook function to ClientConfig like:

```go
PublishResponseHook func(incommingPacket, outgoingPacket *packets.Packet)
```

And this hook will be defined like:
```go
paho.ClientConfig{
    ..snip..

    PublishResponseHook: func(incomingPacket, outgoingPacket *packets.Packet) {
        pb := incomingPacket.(*packets.Publish)
        // do something by using publish's properties

        pa := outgoingPacket.(*packets.Puback)
        if pa.Properties == nil {
            pa.Properties = &packets.Properties{
                User: []packets.User{Key: "hoge", Value: "fuga"},
            }
        }
    },
}
```

Then this hook can be used here:
https://github.com/eclipse/paho.golang/blob/76a1f9a081d13c3e6da7706c7dd3e24c8139d82c/paho/client.go#L315 

like:
```go
c.ClientConfig.PublishResponseHook(pa, &pb)
```

Thanks,

Ryosuke

alsm:
What are you thinking of doing with the hooks? Puback and other such packets are really for the protocol and implementations rather than users.

alsm:
The publish hook is before messages are sent to allow the user of the library to modify them, and currently to provide automatic topic aliasing.

ryskiwt:
I am currently developing an MQTT broker with its own extension specification. I would like to return additional information defined in the extension specification to the MQTT broker using UserProperty in PubAck, PubRec or PubComp messages. 

However, there seems to be no library that has hooks like I proposed. If this library could support such hooks, it would save me a lot of time and effort to develop a client that supports the extension specifications.

alsm:
I've been thinking about this, on the one hand this opens up a whole load of possibilities that weren't previously possible, and I don't think were imagined when the spec was put together, it will be quite possible to create tight coupling between an application and a specific broker.
On the other hand this opens up a whole load of possibilities that weren't previously possible and it might be exciting to allow that to be explored.
If anyone else is reading this issue I'd appreciate other thoughts.

alsm:
Well, as no one else has responded I will have a go at adding this and see how it looks/feels. Not going to guarantee that I won't remove it in the future if I decide I don't like it :)

PR #52, opened 2021-03-25
In-order routing and late ack

fracasula:
Sorry for not creating an issue first but I developed this for the company I work for so I'm creating a draft PR to test the waters. I think it may be more useful to discuss this with the code changes at hand. If it doesn't work for you I can still open an issue first, please let me know.

This PR aims to solve two issues:
1. Message ordering: https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901240
2. Message acknowledgment only after the message processing has completed to ensure re-delivery of "unacked" messages from the server in case of abnormal client termination

In order to avoid blocking the `Incoming()` loop (which would lead to unexpected behaviours such as client disconnections etc...) you are routing the `PUBLISH` packets asynchronously (i.e. by creating a `go` routine: `go c.Router.Route(pb)`). Thus a `go` routine is created for each `PUBLISH` packet, which changes the order of the messages as received from the server. See example in the [Go Playground](https://play.golang.org/p/7ihnnxm7WT7).

The solution I'm proposing is to queue the `PUBLISH` packets in the same order as they are being received in a buffered channel of the size of the "receive maximum" (worst case the buffer size would be `65535`).

We then create a new worker in `Connect()` that listens to the new buffered channel and calls the router without creating a `go` routine. It is up to the user to create a routine in their message handler if they don't want to block. This way the client would give users maximum flexibility when building routers (blocking vs non-blocking APIs).

Since the call to the router in this PR is blocking we can now make sure that the `PUBACK` is sent to the server only after the message handler is done with the message. In case of abnormal termination of the client the `PUBACK` won't be sent to the server which will redeliver the related `PUBLISH` packet (according to "clean start" and "session expiry interval" settings).

This has been tested with running instances of VerneMQ with these settings:
* 10 producers publishing 100k messages each (concurrently)
* a single consumer with receive maximum 65535 and session expiry interval 0xffffffff

The consumer is clearly swamped with messages but given the "receive maximum" setting it receives only 65535 `PUBLISH` packets which fit perfectly in the buffered channel so the `Incoming()` loop is never blocked.

**Testing**

I'm not sure about test coverage here, I'd write an integration test with a real broker running but it doesn't look like you have support for integration tests in this project. Do you think an additional unit test could be useful here? Any pointers?

### DIFF
--- a/paho/client.go
+++ b/paho/client.go
@@ -147,18 +147,23 @@ func NewClient(conf ClientConfig) *Client {
 // returned. Otherwise the failure Connack (if there is one) is returned
 // along with an error indicating the reason for the failure to connect.
 func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
+	if c.Conn == nil {
+		return nil, fmt.Errorf("client connection is nil")
+	}
+
 	cleanup := func() {
+		c.mu.Lock()
+		defer c.mu.Unlock()
+		defer c.workers.Wait()
+
 		select {
 		case <-c.stop:
-			//already shutting down, do nothing
+		//already shutting down, do nothing
 		default:
 			close(c.stop)
 			close(c.publishPackets)
 		}
 		_ = c.Conn.Close()
-	}
-	if c.Conn == nil {
-		return nil, fmt.Errorf("client connection is nil")
 	}
 
 	c.stop = make(chan struct{})
@@ -194,6 +199,7 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
 	c.workers.Add(1)
 	go func() {
 		defer c.workers.Done()
+		defer c.debug.Println("returning from publish packets loop worker")
 		c.routePublishPackets()
 	}()
 
@@ -201,6 +207,7 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
 	c.workers.Add(1)
 	go func() {
 		defer c.workers.Done()
+		defer c.debug.Println("returning from incoming worker")
 		c.Incoming()
 	}()
 
@@ -278,6 +285,7 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
 	c.workers.Add(1)
 	go func() {
 		defer c.workers.Done()
+		defer c.debug.Println("returning from ping handler worker")
 		c.PingHandler.Start(c.Conn, time.Duration(keepalive)*time.Second)
 	}()
 
@@ -285,28 +293,34 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
 }
 
 func (c *Client) routePublishPackets() {
-	for pb := range c.publishPackets {
-		c.Router.Route(pb)
-		switch pb.QoS {
-		case 1:
-			pa := packets.Puback{
-				Properties: &packets.Properties{},
-				PacketID:   pb.PacketID,
-			}
-			c.debug.Println("sending PUBACK")
-			_, err := pa.WriteTo(c.Conn)
-			if err != nil {
-				c.errors.Printf("failed to send PUBACK for %d: %s", pb.PacketID, err)
-			}
-		case 2:
-			pr := packets.Pubrec{
-				Properties: &packets.Properties{},
-				PacketID:   pb.PacketID,
-			}
-			c.debug.Printf("sending PUBREC")
-			_, err := pr.WriteTo(c.Conn)
-			if err != nil {
-				c.errors.Printf("failed to send PUBREC for %d: %s", pb.PacketID, err)
+	for {
+		select {
+		case <-c.stop:
+			c.debug.Println("client stopping, routePublishPackets stopping")
+			return
+		case pb := <-c.publishPackets:
+			c.Router.Route(pb)
+			switch pb.QoS {
+			case 1:
+				pa := packets.Puback{
+					Properties: &packets.Properties{},
+					PacketID:   pb.PacketID,
+				}
+				c.debug.Println("sending PUBACK")
+				_, err := pa.WriteTo(c.Conn)
+				if err != nil {
+					c.errors.Printf("failed to send PUBACK for %d: %s", pb.PacketID, err)
+				}
+			case 2:
+				pr := packets.Pubrec{
+					Properties: &packets.Properties{},
+					PacketID:   pb.PacketID,
+				}
+				c.debug.Printf("sending PUBREC")
+				_, err := pr.WriteTo(c.Conn)
+				if err != nil {
+					c.errors.Printf("failed to send PUBREC for %d: %s", pb.PacketID, err)
+				}
 			}
 		}
 	}
@@ -434,6 +448,8 @@ func (c *Client) Incoming() {
 
 func (c *Client) close() {
 	c.mu.Lock()
+	defer c.mu.Unlock()
+
 	select {
 	case <-c.stop:
 		//already shutting down, do nothing
@@ -444,9 +460,8 @@ func (c *Client) close() {
 	c.debug.Println("client stopped")
 	c.PingHandler.Stop()
 	c.debug.Println("ping stopped")
-	c.Conn.Close()
+	_ = c.Conn.Close()
 	c.debug.Println("conn closed")
-	c.mu.Unlock()
 }
 
 // Error is called to signify that an error situation has occurred, this
--- a/paho/client.go
+++ b/paho/client.go
@@ -296,7 +296,6 @@ func (c *Client) routePublishPackets() {
 	for {
 		select {
 		case <-c.stop:
-			c.debug.Println("client stopping, routePublishPackets stopping")
 			return
 		case pb := <-c.publishPackets:
 			c.Router.Route(pb)
--- a/paho/client.go
+++ b/paho/client.go
@@ -3,6 +3,7 @@ package paho
 import (
 	"context"
 	"fmt"
+	"math"
 	"net"
 	"strings"
 	"sync"
@@ -54,6 +55,7 @@ type (
 		caCtx          *caContext
 		raCtx          *CPContext
 		stop           chan struct{}
+		publishPackets chan *packets.Publish
 		workers        sync.WaitGroup
 		serverProps    CommsProperties
 		clientProps    CommsProperties
@@ -151,14 +153,21 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
 			//already shutting down, do nothing
 		default:
 			close(c.stop)
+			close(c.publishPackets)
 		}
-		c.Conn.Close()
+		_ = c.Conn.Close()
 	}
 	if c.Conn == nil {
 		return nil, fmt.Errorf("client connection is nil")
 	}
 
 	c.stop = make(chan struct{})
+
+	var publishPacketsSize uint16 = math.MaxUint16
+	if cp.Properties != nil && cp.Properties.ReceiveMaximum != nil {
+		publishPacketsSize = *cp.Properties.ReceiveMaximum
+	}
+	c.publishPackets = make(chan *packets.Publish, publishPacketsSize)
 
 	c.debug.Println("connecting")
 	c.mu.Lock()
@@ -180,6 +189,13 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
 			c.clientProps.TopicAliasMaximum = *cp.Properties.TopicAliasMaximum
 		}
 	}
+
+	c.debug.Println("starting publish packets loop")
+	c.workers.Add(1)
+	go func() {
+		defer c.workers.Done()
+		c.routePublishPackets()
+	}()
 
 	c.debug.Println("starting Incoming")
 	c.workers.Add(1)
@@ -268,6 +284,34 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
 	return ca, nil
 }
 
+func (c *Client) routePublishPackets() {
+	for pb := range c.publishPackets {
+		c.Router.Route(pb)
+		switch pb.QoS {
+		case 1:
+			pa := packets.Puback{
+				Properties: &packets.Properties{},
+				PacketID:   pb.PacketID,
+			}
+			c.debug.Println("sending PUBACK")
+			_, err := pa.WriteTo(c.Conn)
+			if err != nil {
+				c.errors.Printf("failed to send PUBACK for %d: %s", pb.PacketID, err)
+			}
+		case 2:
+			pr := packets.Pubrec{
+				Properties: &packets.Properties{},
+				PacketID:   pb.PacketID,
+			}
+			c.debug.Printf("sending PUBREC")
+			_, err := pr.WriteTo(c.Conn)
+			if err != nil {
+				c.errors.Printf("failed to send PUBREC for %d: %s", pb.PacketID, err)
+			}
+		}
+	}
+}
+
 // Incoming is the Client function that reads and handles incoming
 // packets from the server. The function is started as a goroutine
 // from Connect(), it exits when it receives a server initiated
@@ -314,29 +358,7 @@ func (c *Client) Incoming() {
 			case packets.PUBLISH:
 				pb := recv.Content.(*packets.Publish)
 				c.debug.Printf("received QoS%d PUBLISH", pb.QoS)
-				go c.Router.Route(pb)
-				switch pb.QoS {
-				case 1:
-					pa := packets.Puback{
-						Properties: &packets.Properties{},
-						PacketID:   pb.PacketID,
-					}
-					c.debug.Println("sending PUBACK")
-					_, err := pa.WriteTo(c.Conn)
-					if err != nil {
-						c.errors.Printf("failed to send PUBACK for %d: %s", pa.PacketID, err)
-					}
-				case 2:
-					pr := packets.Pubrec{
-						Properties: &packets.Properties{},
-						PacketID:   pb.PacketID,
-					}
-					c.debug.Printf("sending PUBREC")
-					_, err := pr.WriteTo(c.Conn)
-					if err != nil {
-						c.errors.Printf("failed to send PUBREC for %d: %s", pr.PacketID, err)
-					}
-				}
+				c.publishPackets <- pb
 			case packets.PUBACK, packets.PUBCOMP, packets.SUBACK, packets.UNSUBACK:
 				c.debug.Printf("received %s packet with id %d", recv.PacketType(), recv.PacketID())
 				if cpCtx := c.MIDs.Get(recv.PacketID()); cpCtx != nil {
@@ -417,6 +439,7 @@ func (c *Client) close() {
 		//already shutting down, do nothing
 	default:
 		close(c.stop)
+		close(c.publishPackets)
 	}
 	c.debug.Println("client stopped")
 	c.PingHandler.Stop()
--- a/paho/client.go
+++ b/paho/client.go
@@ -297,7 +297,10 @@ func (c *Client) routePublishPackets() {
 		select {
 		case <-c.stop:
 			return
-		case pb := <-c.publishPackets:
+		case pb, open := <-c.publishPackets:
+			if !open {
+				return
+			}
 			c.Router.Route(pb)
 			switch pb.QoS {
 			case 1:
--- a/paho/client_test.go
+++ b/paho/client_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"reflect"
 	"sync"
 	"testing"
 	"time"
@@ -429,15 +430,17 @@ func TestClientReceiveAndAckInOrder(t *testing.T) {
 	wg.Wait()
 
 	require.Equal(t, expectedPublishPackets, actualPublishPackets)
-	require.Len(t, ts.receivedPubacks, 3)
-	require.Equal(t,
-		[]*packets.Puback{
-			{PacketID: 1, ReasonCode: 0, Properties: &packets.Properties{}},
-			{PacketID: 2, ReasonCode: 0, Properties: &packets.Properties{}},
-			{PacketID: 3, ReasonCode: 0, Properties: &packets.Properties{}},
-		},
-		ts.receivedPubacks,
-	)
+	require.Len(t, ts.ReceivedPubacks(), 3)
+	require.Eventually(t, func() bool {
+		return reflect.DeepEqual(
+			[]*packets.Puback{
+				{PacketID: 1, ReasonCode: 0, Properties: &packets.Properties{}},
+				{PacketID: 2, ReasonCode: 0, Properties: &packets.Properties{}},
+				{PacketID: 3, ReasonCode: 0, Properties: &packets.Properties{}},
+			},
+			ts.ReceivedPubacks(),
+		)
+	}, time.Second, 10*time.Millisecond)
 }
 
 func TestReceiveServerDisconnect(t *testing.T) {
--- a/paho/client_test.go
+++ b/paho/client_test.go
@@ -2,8 +2,10 @@ package paho
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"os"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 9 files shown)


alsm:
Thanks for this, I'll take a closer look tomorrow.

fracasula:
@alsm I also wanted to discuss the possibility to let the user acknowledge messages when they want. One possible solution would be to add an `Ack()` method to the `Client` so instead of:

```go
func (c *Client) routePublishPackets() {
	for pb := range c.publishPackets {
		c.Router.Route(pb)
		switch pb.QoS {
		case 1:
			pa := packets.Puback{
				Properties: &packets.Properties{},
				PacketID:   pb.PacketID,
			}
			c.debug.Println("sending PUBACK")
			_, err := pa.WriteTo(c.Conn)
			if err != nil {
				c.errors.Printf("failed to send PUBACK for %d: %s", pb.PacketID, err)
			}
		case 2:
			pr := packets.Pubrec{
				Properties: &packets.Properties{},
				PacketID:   pb.PacketID,
			}
			c.debug.Printf("sending PUBREC")
			_, err := pr.WriteTo(c.Conn)
			if err != nil {
				c.errors.Printf("failed to send PUBREC for %d: %s", pb.PacketID, err)
			}
		}
	}
}
```

We could do:

```go
func (c *Client) routePublishPackets() {
	for pb := range c.publishPackets {
		c.Router.Route(pb)
	}
}

// Ack is used to send a PUBACK back to the server for QoS 1 and 2 messages
func (c *Client) Ack(pb *packets.Publish) error {
	switch pb.QoS {
	case 1:
		pa := packets.Puback{
			Properties: &packets.Properties{},
			PacketID:   pb.PacketID,
		}
		c.debug.Println("sending PUBACK")
		_, err := pa.WriteTo(c.Conn)
		if err != nil {
			c.errors.Printf("failed to send PUBACK for %d: %s", pb.PacketID, err)
		}
		return err
	case 2:
		pr := packets.Pubrec{
			Properties: &packets.Properties{},
			PacketID:   pb.PacketID,
		}
		c.debug.Printf("sending PUBREC")
		_, err := pr.WriteTo(c.Conn)
		if err != nil {
			c.errors.Printf("failed to send PUBREC for %d: %s", pb.PacketID, err)
		}
		return err
	}
	return nil
}
```

This way users can still create go routines to process messages concurrently but still in order according to their own business logic.

For example, let's say that a consumer is subscribed to `+/temperature` where the `+` wildcard is the client name e.g. `sensor-1/temperature`, `sensor-2/temperature` etcetera.

In the client the messages will come in order thanks to the buffered channel but then the user router can still spawn one go routine per sensor, one for `sensor-1` and one for `sensor-2` (or have dedicated channels per sensor) and still be able to process the messages in order for all sensors without risking of acknowledging a message that we are not done with.

fracasula:
Yes I can see how difficult it could become to maintain the order of acknowledgments if we were to expose an `Ack()` method. I will put more thought on this to see if I can think of something, in the meantime thank you.

Going back to the original proposal though, if we block on the acking part we're not breaking anything, we're still following the specs and we're sending the acks in the order they were received by the client. We are just letting the users decide whether it is the case for them to block or not in their router. Performance wise it's always possible to decrease the `ReceiveMaximum` window and avoid prefetching too many messages if you're planning to block for example. There are many things that can be done on the broker side as well.

One more thing about performance. I think it matters but on a different level and that is a combination of application (not client) design and client/server settings. First of all we have to make sure that we meet the requirements (e.g. avoid message loss) then if performance becomes an issue one should either relax the requirements or use a different topology. Shared-subscriptions and clustering are examples of how a topology can change to meet the performance needs.

My problem with the current approach is that the ack/routing behaviour is hard-coded in the client just because it might cause performance issues in some use cases? (with what brokers and what settings btw? we've been doing this in production for months with VMQ and we are satisfied with the performance, we got 26k agents connected simultaneously sending messages).

Also, I couldn't find any spec saying that the `PUBACK` should be sent immediately after a `PUBLISH` with no delay. Is there such a spec? If yes please let me know.

If you don't like this approach, do you have any other idea about how we can avoid message loss on ungraceful client termination?

alsm:
You're of course right that as currently written your PR doesn't make a huge difference to the way the client runs, and I wrote the client such that we could have pluggable router implementations, unlike the v3 client we aren't tied to a single style. I guess the consideration is what should be the default, what is the least surprising behaviour. Clearly out of order reception is not it so I definitely want the ordering you offer. On the delay to responding to a PUBLISH, this is in the spec in section 3.3.4;

> The Client MUST NOT delay the sending of any packets other than PUBLISH packets due to having sent Receive Maximum PUBLISH packets without receiving acknowledgements for them [MQTT-3.3.4-8]. The value of Receive Maximum applies only to the current Network Connection.

This doesn't actually say that they should be sent immediately, but is waiting for a message handler to return a delay? I would say that it is as that interaction between the client library and an application is outside the scope of the spec. Then again most message handlers will return fairly quickly and how would the server know any different? I can see the attraction of doing it this way, it negates the need for any incoming persistence store for the client. I'm leaning towards accepting it as it's a beneficial change I just want to think a little more on it and talk it over with some other people.

dambrisco:
> I am against allowing applications to have any control over the acking of messages, I let this into the v3 client and I don't think it was really beneficial, there are couple of reasons. Firstly the contract as regards handling messages is between the server and the client, once the client has invoked a handler for a message its responsibility ends, if the handler subsequently crashses, blocks, loses the message that's unfortunate but not really the client's issue. Secondly it's almost certainly going to cause problems with spec compliance and performance.

Just wanted to chime in that as active users of the v3 client, we actually _rely_ on the ability to `ack` messages ourselves. Don't mean to derail the thread or insinuate that supporting such a feature is necessary, but fwiw our perspective is that paho is simply a _library_ we use to write our _own_ client.

fracasula:
@alsm thanks for getting back to me, all this has been very informative for me so far :bow:  

In regards to the spec you quote, I'm not sure I understand them fully, let's see if you can help me out a bit:

> The Client MUST NOT delay the sending of any packets other than PUBLISH packets due to having sent Receive Maximum PUBLISH packets without receiving acknowledgements for them [MQTT-3.3.4-8]. The value of Receive Maximum applies only to the current Network Connection.

Am I correct in assuming that if the Client is merely consuming and not publishing anything (thus not sending any `PUBLISH` packets) then the condition does not apply? I'm referring to the fact that it must not delay packets due to having sent `PUBLISH` packets, so if it doesn't send them then the rule doesn't apply? Not sure it's that simple though but if that is the case then that's kind of the thing that I'm trying to build, an MQTT bridge to Apache Pulsar.

Anyway, sorry for the little digression, I just wanted to make sure that I'm getting this right. Thanks again and in the meantime I'll wait for you to get back to us.

alsm:
Discussing this with @icraggs I am happy for this to go in, it matches the way the paho C client handles this situation. Looking at the code I wonder if `routePublishPackets()` should be doing a select on c.stop as well, if the channel has packets in but the client is disconnected do we want to continue to process the the already received messages? Unacked messages will be resent by the server when the client reconnects. I've just noticed I'm not actually waiting on the client waitgroup to return, but I should and when I add that waiting for processing of received messages to complete could block notification that the client is disconnected.

alsm:
> Am I correct in assuming that if the Client is merely consuming and not publishing anything (thus not sending any `PUBLISH` packets) then the condition does not apply? I'm referring to the fact that it must not delay packets due to having sent `PUBLISH` packets, so if it doesn't send them then the rule doesn't apply? Not sure it's that simple though but if that is the case then that's kind of the thing that I'm trying to build, an MQTT bridge to Apache Pulsar.
> 
> Anyway, sorry for the little digression, I just wanted to make sure that I'm getting this right. Thanks again and in the meantime I'll wait for you to get back to us.

It seems not everyone reads this section the same way :) I read it as "The client must not delay the sending of any packets, other than publish packets that want to be sent but cannot due to there being receive maximum publishes already inflight with the server."
@icraggs - "I read that spec statement as referring to the receive maximum limit, making it clear that it applies to PUBLISH packets only, not PUBACKs, PUBRECs, PUBRELs or PUBCOMPs. I do think however there is a question of ordering, that the ack for a message shouldn't be delayed so long that the ack for a subsequently received PUBLISH is sent first, at least not routinely."

With Ian's reading it doesn't seem to be applicable to your case at all.

fracasula:
> Discussing this with @icraggs I am happy for this to go in, it matches the way the paho C client handles this situation.

That's great to hear @alsm!

> Looking at the code I wonder if `routePublishPackets()` should be doing a select on c.stop as well

It makes sense yeah, I pushed some more changes that make the routine exit on `<-c.stop`.

> I've just noticed I'm not actually waiting on the client waitgroup to return, but I should and when I add that waiting for processing of received messages to complete could block notification that the client is disconnected.

I took the liberty of adding a `defer c.workers.Wait()` in the `cleanup` lambda function that is defined in the `Connect()` method. Let me know if it makes sense. 

I also tried to add the wait in the `close()` method but that can lead to a deadlock when the `Incoming` loop comes across an error for example. In that case `c.Error()` is called which calls `c.close()`. So if `c.close()` where to wait on the workers it will never finish given that one of the workers is `Incoming()`.

My question is, where do you think it makes more sense to wait for the workers? I think in `close()`, if you agree we might need to rethink a few things in order to avoid the deadlock though. Anyway, it should be done in a different PR I guess.

**Note:** please have a look at the new test, I also added some stuff to the test server to spy on the `PUBACK` and `PUBREC` packets sent by the client.

paho/client.go
```diff
@@ -145,21 +147,33 @@ func NewClient(conf ClientConfig) *Client {
 // returned. Otherwise the failure Connack (if there is one) is returned
 // along with an error indicating the reason for the failure to connect.
 func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
+	if c.Conn == nil {
```
fracasula, line 150: I moved this `if` to the top because if we return right away then there's no need to initialize `cleanup`.

paho/client_test.go
```diff
@@ -353,6 +366,80 @@ func TestClientReceiveQoS2(t *testing.T) {
 	<-rChan
 }
 
+func TestClientReceiveAndAckInOrder(t *testing.T) {
```
fracasula, line 369: I used a slightly different approach in this test, I hope you don't mind.
Instead of starting routines manually and initialize unexported parts of the clients, I'm initializing the client in a way that is closer to how a user would do it (given that they can't just do `c.stop = make(chan struct{})`). I find it less error prone, let me know what you think.

paho/client.go
```diff
@@ -262,12 +285,49 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
...
+func (c *Client) routePublishPackets() {
+	for {
+		select {
+		case <-c.stop:
+			return
+		case pb, open := <-c.publishPackets:
+			if !open {
+				return
+			}
+			c.Router.Route(pb)
```
MattBrittan, line 304: One potential concern with this is that the processing runs on a single thread. This could be an issue where the handler is storing the message to a remote database (or other high latency process).  I cannot see how you can process multiple messages simultaneously (because `routePublishPackets` will block until `Route` completes). Allowing for this may be a bit complicated due to the spec requirement that acknowledgments be sent "in the order in which the corresponding PUBLISH packets were received". I don't have a solution to this right now (will think about it) but wanted to raise it (would be great to get the interface right on the first attempt!). Great PR by the way!.

fracasula: Hey, thanks! If you want you can just avoid blocking inside your router:

```go
c := NewClient(ClientConfig{
	Conn: ts.ClientConn(),
	Router: NewSingleHandlerRouter(func(p *Publish) {
		go doSomething(p)
	}),
})
```

This way at least you have both options, blocking and not-blocking. Plus you can handle that on your side. Of course, if you do that, you lose the benefit of having the message redelivered by the broker in case you weren't able to complete your "side-effect" operation successfully.

MattBrittan: Sorry, I probably need to add a bit of context to this. I'm starting work on adding persistence and, initially anyway, we  (discussed with @alsm) plan on implementing outgoing persistence only (on the basis that, with your change, the broker effectively handles inbound persistence - realise there are QOS2 issues but...). On that basis I was keen to have a way of processing incoming publish packets, in parallel, whilst retaining control of the ACK (meaning your suggestion of `go doSomething(p)` is not quite sufficient).
The idea I'm currently mulling is to expand the approach you have taken and send ALL packets (both sent and received) via a channel to a `router` which can take whatever action it wants before taking action (which may be just sending a packet, calling a handler, sending a response to a received packet etc (need to write this up in more detail but will probably do this as a test implementation because it's probably easier to see what works that way). There would be a generic implementation of this which just sends the standard response (probably with a delay to allow the handler to run in the case of a publish packet) but users would be able to replace the `router` thus enabling the implementation of persistence via a wrapper (rather than adding significant complexity to the main `paho` package) and whatever other message handling needs users may have (e.g. issue #40 would be covered).
Very keen to hear your thoughts on this (it's probably going to take a few iterations to get right but I figured it was worth mentioning because it may influence this code and I believe you have more experience using it than most!).

fracasula: Yes I understand your point and I think it's an interesting approach. I was hoping I could get things done in different PRs though. If that doesn't work of course it's OK, just expressing my preference. That's why I proposed the `go doSomething(p)` in the router because that way we can still manage to merge this PR and potentially have exactly what we have now on `master` (with the difference that we go async in the router rather than in the client) plus the in-order delivery if you avoid blocking. So with approach we would still be quite backwards compatible and manage to merge an improvement.

Then I had another idea for increasing performance and I spent some time last week testing my approach, which I haven't proposed yet (I was planning to do so in another PR).

My use case is a bridge from the MQTT broker to Apache Pulsar. As you can imagine doing one message at a time leads to very poor results, especially considering that Pulsar is distributed and there's consensus involved as well so multiple nodes need to acknowledge before they get back with an ack to the client. On my local machine we're talking ~675 messages per second (quite poor).

So I started experimenting with the idea of batching. If instead of pushing a single message at a time I buffer them and then write them as a batch to Pulsar (similar to a MongoDB [InsertMany](https://github.com/mongodb/mongo-go-driver/blob/457710110ff38c16006d73a656d108d432f935ca/benchmark/multi.go#L43)), then on the same machine I get more than 100000 messages done per second (vs 675, quite an improvement!).

At this point acknowledging in order on the MQTT side would be quite simple because we just buffered, say the first 1000 messages, then stop, write the batch, ack them in order and carry on.

With this in mind I started thinking of a way to give users an API to avoid messing up with the ack order. And then I thought of cumulative acking. Here's the idea:
* the client by default behaves like in this PR
* it has a configurable attribute `EnableCumulativeAcking`
* if `EnableCumulativeAcking` is `true` then the client doesn't ack automatically and keeps track of `packetID` and QoS of every packet it comes across and stores them in order in a slice (memory footprint is minimal, max it could be 65535 times a struct with just those two attributes (`uint16` and `int` if I remember well)
* then we expose a `CumulativeAck()` method that, given a `packetID`, it looks for it in the slice and starts acking from the beginning till that packet that has found

With an API like that, once you write your batch (which performs even better then going in parallel in most of the benchmarks I wrote), you just `CumulativeAck()` the last one and then the client does the magic for you with what I think it's little complexity.

Of course this might require some thinking when designing your application since things like batching usually can have a target only, for example a single topic on Pulsar or a single collection on MongoDB. Thus the need to make sure on your side that whatever comes out of a topic should go in a single place (e.g. collection/topic) only in order to get the most out of it.

Main advantages:
* keeps the default as is (for backwards compatibility)
* you can still ack a single message with `CumulativeAck()`
* given the method name, the requirement of acking in order becomes pretty clear for early users as well
* performs quite nicely
* decreases complexity vs a pure in-parallel approach (given the requirements). say that you process 100 messages in parallel but only one doesn't succeed, you hold back the other 99 for as long as a timeout. there might be retries involved and if you decide to ack it anyway even if you can't apply the side effect you might lose the message. with batching it's all or none, easier to debug and timeouts would add pressure mostly on the broker rather than in the client (and the broker usually has plenty more resources than a client and hopefully clustering which should help redistribute load). if faster processing is needed then shared subscriptions can be used with the advantage of keeping a single client as simple as possible.

Anyway, sorry for the long post 😅 
I just wanted to share some thoughts, performance is indeed quite important for me as well.

Next week I'll try to review the issue you linked and I'll put more thought on your approach. If you have some code to share please do, I tend to reason better with implementations. The way you describe it seems legit 👍

MattBrittan: Thanks very much for the detailed explanation! My (slight) concern was that this change does alter the API slightly (moving from `go c.Router.Route(pb)` to `c.Router.Route(pb)` which would require end users to alter their code if they wish to avoid sequential processing (a small change, and we are pre 1.0 so to be expected, but this does provide the opportunity to make a larger change that could simplify things going forward).  I'm currently still on 3.1 because persistence is a requirement for me so I cannot really comment too much and, for the avoidance of doubt, have no objections to this PR going ahead as-is (but perhaps consider the thoughts below).

Fleshing out my concern a bit:
 * If users wish to maintain the current functionality (multiple simultaneous processing of inbound messages) they will need to change their code (possibly not a breaking change but could have a major impact in high volume uses).
 * It does not provide a way to process multiple messages simultaneously **AND** control the ACK.
 * If you decide you do not want to acknowledge a packet the only option I can see is to terminate the MQTT connection before returning (and that is problematic because `routePublishPackets` will block so you could not wait for all go routines to exit). 
 * A change would be needed if inbound persistence is implemented in the future (there is no way to say "don't send an ACK" at all).

Ref your proposed "API to avoid messing up with the ack order"; I do wonder if this ends up over complicating things within the `paho` package (I've made quite a few changes to the v3 client and really like the simplicity of the v5 client!). My feeling is that anyone who is batching messages should be fairly competent and could take care of ACK order themselves (or they may choose to ignore this requirement in the spec; my guess is that most brokers would handle out of order ACK's whatever the spec says!).
I wonder if, rather than passing a `Publish` to the handler we passed something like (note: somewhat off the top of my head; I have not implemented this!):

```
// note: In order to maintain compliance with the MQTT speck `Ack` must be called in the order in which packets are received.
type PublishWithAck {
   *Publish
      
   defaultAck *io.WriterTo  // Possibly not needed but would make Ack function simpler!
}

// Ack - Acknowledge the Publish packet 
// Passing a nil will result in the default acknowledgement being sent. Might be important for the user to know if the ACK could not be sent (this has been done with issue #40 in mind; I'm thinking that longer term it will be possible to request all packets in this way)
func (p *PublishWithAck) Ack(io.WriterTo) err {
   // todo... Need to ensure only one ACK is sent (so may need a mutex etc)
 }
```

As an example of how this could be used the following would be needed to maintain compatibility with the current functionality:

```
Router: NewSingleHandlerRouter(func(p *PublishWithAck) {
		go doSomething(p.Publish)
		p.Ack(nil)
	}),
```

I believe that this would support your batching use-case (you could add the packets to a slice and once it hits a predetermined threshold send them all to Pulsar and then call the `.Ack` functions in order). If we wanted to simplify things further for the user then a configurable attribute `DisableAutoPublishAck` could be added (so unless this is called the functionality is pretty much as now except that the router is called sequentially and passed a different type).  Note: This is kind-of similar to the 3.1 client (except that does not support turning off the Auto Ack so it's pretty useless there!).

fracasula: Yes I agree it's not optimal but given the MQTT specs we are given I'm trying to split the work in a reasonable way.
At the moment this client isn't guaranteeing in-order delivery and with this change, if we consider what we're merging against, we fix that and keep the rest relatively the same. As you mention I wouldn't block this PR but all the things you say should be addressed indeed.

The batching idea comes mostly from @alsm concerns expressed [here](https://github.com/eclipse/paho.golang/pull/52#issuecomment-810216683). My first proposition was indeed just an `Ack()` method and then let the user handle the different cases.

Also, as you mention, there are MQTT brokers that do not require acknowledgments in order (i.e. VerneMQ).

If you're OK with this I say we continue this discussion in a new dedicated issue where we try to tackle the acknowledgements. I don't mind creating an issue myself, please let me know what you think.

As a sidenote, I'll go as far as to say that we should try to understand why that requirement was defined in MQTT in the first place and see if we can change it in a new version. Without the requirement of acknowledging in order and with in-order delivery guaranteed over shared-subscriptions, the MQTT protocol would be able to serve a lot more use cases more efficiently (for example EMQX does the in-order delivery over shared-subscriptions by routing via a hash of the topic or the client ID, very similar to what Kafka and Pulsar do with key partitioning).

Thanks!

MattBrittan: Thanks Francesco,
Thinking about this further I can implement the changes mentioned as an option with minimal impact to users so no issues there.
Ref why it was done this way [see this](https://issues.oasis-open.org/browse/MQTT-71). Basically I would see the rationale as something along the lines of "A QOS2 message should only be forwarded when PUBCOMP has been sent and, thus, PUBREC and PUBREL must be sent in order". In reality this does not really impact the end recipient if messages are passed on when the PUB arrives (as happens in this client). In response to @alsm's question re this Ian Craggs commented 

>I do think however there is a question of ordering, that the ack for a message shouldn't be delayed so long that the ack for a subsequently received PUBLISH is sent first, at least not routinely.

so it does not appear that he feels the ordering essential...

Matt

fracasula: @MattBrittan As I said before I'm all for giving the users the ability to `Ack()` themselves. Even with the API that you're proposing though (i.e. `PublishWithAck`), it's easy to mess up with the order. I'm cool with it though because I think the user should be able to handle that. So as long as we document it properly and note that some brokers do not need them in order after all then I'm cool with it (I checked with the EMQX guys and at least for now they don't care about the order as well).

I can also try something if you guys think it is required in this PR after all, but before doing any work I would like to hear from @alsm, which has been silent for a bit now. I think it'd be beneficial if we can all agree on this before doing the changes.

Another option, as I proposed [here](https://github.com/eclipse/paho.golang/pull/52#issuecomment-810158305), is to have a `func (c *Client) Ack(pb *packets.Publish) error` which seems to be a fairly common approach amongst clients.

A third option, is to have the router take a 2nd parameter which would be a lambda function to ack that packet:

```go
case packets.PUBLISH:
	pb := recv.Content.(*packets.Publish)
	ack := func() error {
		switch pb.QoS {
		case 1:
			pa := packets.Puback{
				Properties: &packets.Properties{},
				PacketID:   pb.PacketID,
			}
			return c.write(ctx, &pa)
		case 2:
			pr := packets.Pubrec{
				Properties: &packets.Properties{},
				PacketID:   pb.PacketID,
			}
			return c.write(ctx, &pr)
		}

		return nil
	}

	if c.Router != nil {
		go c.Router.Route(pb, ack)
	} else {
		_ = ack()
	}
```

My preference for now would be to have the `Ack()` in the client and leave the packets as dumb as possible. Something that I like about the packets, as they are now, is that you can safely pass them by value as well, which is a plus for me. Sometimes I prefer a copy in the stack rather than in the heap which might add pressure on the garbage collector. I haven't done any escape analysys though to verify the impact but it's something that we should keep in mind in case we end up putting stuff inside packets that can't be copied just because, in order for them to expose an `.Ack()` method, they end up having pointers to mutexes, client, connections etc...

Thoughts? :slightly_smiling_face:

fracasula: As a sidenote, I'm still not sure I get the reason why clients should send the packets in order. I also don't fully understand the reason why QoS2 requires basically two acknowledgements when other protocols do exactly-once semantics with a single ack. Perhaps it's all built around the assumptions that packets shouldn't be delayed so a single `PUBACK` doesn't give the server the assurance that the client was actually done with the message thus the double ack (`PUBREC`+`PUBREL`) of the QoS2? :thinking:

Do you have any insights on this?

MattBrittan: @fracasula I'll raise a new issue re this because I think manual handling of `ACK` is outside the scope of your original PR and the discussion is likely to get lost if it remains here. As mentioned before I have no issues with this PR going forward but will leave this comment unresolved because, as you say, input from @alsm would be great.

alsm: I'd been away so just catching up on all this...woah 😄 
Initial thoughts;
As a receiving client it should make no difference to the server the order in which I ack the incoming messages, for QoS1 it just means the packet id can be reused, for QoS2 most client shortcut and pass the message to the application on PUBREC rather than PUBCOMP. The only thing here would be if a server doesn't consider a message sent to it to be "published" until the client has sent the PUBREL, but that doesn't have to matter here as we're looking at receiving.
As is though this PR addresses the issue of ordered processing of messages so I'm happy for it to proceed as is and we can look at the acking and performance issues in another issue.

fracasula:
@alsm is there anything else that you think needs to be done here? any thoughts on the test coverage? I ran some integration tests as well on my end, the rest was the unit tests you see in this PR.

alsm:
No, It's all good was just checking we were all done with it

Issue #54, opened 2021-05-27
struck on Connection timeout error


golthitarun:
**Describe the bug**
When connection timeout occurs during client connection the code is blocked in cleanup method. 

I was able to make it run by removing the sync mutex lock code added here recently
https://github.com/eclipse/paho.golang/blob/6f81099163c2cbbe7862b98e28062baca74cd275/paho/client.go#L155

**To reproduce**
Try to reproduce connection timeout in connect method

**Debug output**
Not seeing any output, during debug I observed the code being struck in gopark method of go. https://golang.org/pkg/runtime/?m=all#gopark

**Expected behaviour**
An error response

**Software used:**

**Additional context**

golthitarun:
@fracasula looks like you added that change recently, can you help me look into it.

MattBrittan:
Looks like the issue is due to the use of `defer` [here](https://github.com/eclipse/paho.golang/blob/6f81099163c2cbbe7862b98e28062baca74cd275/paho/client.go#L178):

```
c.mu.Lock()
defer c.mu.Unlock()
```

and then calling `cleanup` within the same function (e.g. [here](https://github.com/eclipse/paho.golang/blob/6f81099163c2cbbe7862b98e28062baca74cd275/paho/client.go#L228)). This means that `cleanup` will be called before the deferred `c.mu.Unlock()` and thus will [block here](https://github.com/eclipse/paho.golang/blob/6f81099163c2cbbe7862b98e28062baca74cd275/paho/client.go#L155).

As `cleanup` is only called when `Connect()` already has `mu` locked the lock within it seems unnecessary.

golthitarun:
Looks like there is also issue with defering waitgroup wait method in cleanup method, Its basically struck waiting. 

I added the defer waitgroup.Wait() to the start of connect method and it was working for me. I have the change in my local commit here -> https://github.com/golthitarun/paho.golang/commit/324831c52b037f47732760205013b09aa662aaec

I can create a PR if this looks good, lemme know

fracasula:
I was able to replicate with a simple test and found 2 issues so far:
1. we get a `panic` when the `ctx` is canceled before `Connect` is called, this is the affected code: https://github.com/eclipse/paho.golang/blob/master/paho/client.go#L236
2. even if we remove the first 3 lines from the `cleanup` function and add a `defer c.workers.Wait()` in `Connect` we get a deadlock, the worker that doesn't terminate seems to be the `PingHandler`

[...]

PR #55, opened 2021-05-28
Deadlock

fracasula:
This PR is to address the deadlock on cleanups, see [here](https://github.com/eclipse/paho.golang/issues/54).

It also addresses another bug that I found related to canceled contexts. It was causing panics due to the usual nil pointer dereference (e.g. `cap *packets.Connack` was passed as `nil` [here](https://github.com/eclipse/paho.golang/blob/6f81099163c2cbbe7862b98e28062baca74cd275/paho/client.go#L244) causing the `panic`).

To fix the deadlocks I'm proposing an idempotent `close()`. This way we don't have to worry about calling it multiple times from different places. Once called the method checks whether it was being closed already or not.

Even though not all calls to `close()` are blocking ones due to this approach, the code should still offer a blocking API to the user considering how intertwined are the methods. `close()` is unexported and always called by a more important method which  actually blocks.

I added a test to verify the correct cleanup behaviour and to test the context cancelation.

**Note:** I hope you don't mind the Makefile, I added a recipe to always run the tests with the race enabled.

Fixes: https://github.com/eclipse/paho.golang/issues/54

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,4 @@
+.PHONY: test
+test:
+	go test -race ./packets/ -v -count 1
+	go test -race ./paho/ -v -count 1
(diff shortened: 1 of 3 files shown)


fracasula:
I think this would require further testing, I'm not sure of this approach. I'll try to see if I can spend some more time on this next week. In the meantime, suggestions are of course welcome.

PR #56, opened 2021-06-03
Deadlock + minor bugfixes

fracasula:
This PR is to address the deadlock on cleanups, see [here](https://github.com/eclipse/paho.golang/issues/54).

It also addresses another bug that I found related to canceled contexts. It was causing panics due to the usual nil pointer dereference (e.g. `cap *packets.Connack` was passed as `nil` [here](https://github.com/eclipse/paho.golang/blob/6f81099163c2cbbe7862b98e28062baca74cd275/paho/client.go#L244) causing the `panic`).

### Changes
* wait for `CONNACK` without starting `incoming` worker
  * this helped with the removal of the deadlock
* handle all context errors
* move all workers at the end of `Connect` when all operations have completed successfully
* making `incoming` and `error` methods unexported
  * this helped me restrict the context where the lock could be acquired by the user calling a method directly (e.g. `Authenticate`, `Connect`...)
  * it makes more clear that the user isn't supposed to call those methods
  * the methods can still be called internally in tests
*  new `TestCleanup` to test for other context errors (e.g. canceled context) and to test for the deadlock
   * I wrote the test before making the changes and it was indeed locking forever. once I could replicate the deadlock I worked on the fix I'm proposing with this PR
* new `TestDisconnect` which tests that `Disconnect()` doesn't lock and that it's also possible to disconnect multiple times without locking
* new `TestCloseDeadlock` which calls `close()` and `Disconnect` concurrently multiple times to see if a deadlock occurs
* `Makefile` with recipe to automate execution of all tests (with the `-race` flag)

Fixes: https://github.com/eclipse/paho.golang/issues/54

### DIFF
--- a/paho/client.go
+++ b/paho/client.go
@@ -778,10 +778,11 @@ func (c *Client) publishQoS12(ctx context.Context, pb *packets.Publish) (*Publis
 func (c *Client) Disconnect(d *Disconnect) error {
 	c.debug.Println("disconnecting")
 	c.mu.Lock()
-	defer c.mu.Unlock()
-	defer func() { _ = c.Conn.Close() }()
-
 	_, err := d.Packet().WriteTo(c.Conn)
+	c.mu.Unlock()
+
+	c.close()
+	c.workers.Wait() // wait anyway in case close was already called
 
 	return err
 }
--- a/paho/client.go
+++ b/paho/client.go
@@ -49,9 +49,7 @@ type (
 	Client struct {
 		mu sync.Mutex
 		ClientConfig
-		// caCtx is used for synchronously handling the connect/connack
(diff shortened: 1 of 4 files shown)


paho/client.go
```diff
@@ -52,7 +52,6 @@ type (
 		// caCtx is used for synchronously handling the connect/connack
```
alsm, line 52: Should fix up this comment if caCtx is going away

fracasula: Good catch, fixed :+1:

alsm:
Cool, everything started out as exported with the view of moving it to unexported as required by use rather than trying to predict what would or wouldn't be needed. This seems an appropriate time to move those two.

alsm:
Thanks for your continued contribution to this project.

Issue #53, opened 2021-04-21
Control PUBACK/PUBREC transmission


MattBrittan:
I would like a way to:

* Prevent a `PUBACK`/`PUBREC` from being sent if something goes wrong whilst the message is being processed in my code  (e.g. a service used to persist the message fails).
* Delay the sending of the `PUBACK`/`PUBREC`  so that data can be committed to storage in batches with all relevant messages being acknowledged when the batch has been successfully processed.
* Potentially send a customised `PUBACK` (adding user properties or similar as discussed in issue #40; this only goes part way towards delivering on that request). This would also enable the use of reason codes such as `Topic Name invalid`.
* Potentially use this technique instead of implementing persistence on inbound messages (effectively using the broker as a store). 

This issue comes out of a discussion on PR #52; that PR ensures that the message router is called in message order and that the relevant `PUBACK`/`PUBREC` is returned in order and delays the sending of the `PUBACK`/`PUBREC` until the router returns. The discussion was getting fairly long and is really separate from the PR so I felt it best to open a new issue to allow discussion on this point (it's something others may be interested in).

Relevant issues raised in the V3 Client:

* [Support custom ACK function, check handler function status](https://github.com/eclipse/paho.mqtt.golang/issues/396)
* [Auto-ack of messages in router.go renders API call mqtt.Message.Ack() useless](https://github.com/eclipse/paho.mqtt.golang/issues/459)

**The solution I'd like**

Add an additional parameter to `Router.Route(*packets.Publish)`; `Router.Route(*packets.Publish, func Ack(io.WriterTo) err)`. The `Ack` function will send the default acknowledgment (if `WriterTo` is nil) or use the passed in custom `WriterTo`. If the `Ack` function is not called then no `PUBACK`/`PUBREC` will be sent (and the broker would resend the message at some point). Note that at QOS0 the `ACK` function would not do anything.

The benefits of this solution are:

* Its a clear change to the `Router` interface so obvious that action is needed (some of the other options could lead to the behaviour being hidden). The default router could be changed to call `Ack(nil)` which would mean that most users would not notice the change.
* The `Publish` packet remains simple and pass by value is not an issue (a mutex may be needed so using a wrapper around the publish packet might create issues for users who copy the struct).
* It meets all of the requirements noted above.

**Alternatives considered**

[...]

PR #57, opened 2021-06-07
Manual acknowledgments

fracasula:
This PR is to address this issue: https://github.com/eclipse/paho.golang/issues/53

We want the ability to manually control the PUBACK/PUBREC transmission in order to avoid message loss.

The idea is to acknowledge a message only after it has been processed successfully by a client.

### Changes
* the default behaviour is not changed, acks will be sent in order automatically by the client, right after routing
* if the users want to manually enable acknowledgments they just have to set `EnableManualAcknowledgment` to `true` and then call `Ack()` in any order. the client will take care of sending the acks to the server back in right order.
  * this is something I've seen implemented in other MQTT clients (e.g. HiveMQ Java client)
* a background routine will periodically try to flush the acks in order, by acquiring a lock, as often as configured via the `SendAcksInterval` attribute

### Sidenotes
* this code has been running for several days with a few clients in our staging environment with no issues (with manual acks enabled)
* we're planning to go live in production with this branch in 1-2 weeks here at [Netdata](https://github.com/netdata/netdata) which should give us a certain degree of certainty that it works and performs well. of course we'll keep monitoring it afterwards as well.

### DIFF
--- a/packets/publish.go
+++ b/packets/publish.go
@@ -50,7 +50,7 @@ func (p *Publish) Buffers() net.Buffers {
 	var b bytes.Buffer
 	writeString(p.Topic, &b)
 	if p.QoS > 0 {
-		writeUint16(p.PacketID, &b)
+		_ = writeUint16(p.PacketID, &b)
 	}
 	idvp := p.Properties.Pack(PUBLISH)
 	encodeVBIdirect(len(idvp), &b)
--- a/paho/acks_tracker.go
+++ b/paho/acks_tracker.go
@@ -1,0 +1,79 @@
+package paho
+
+import (
+	"errors"
+	"sync"
+
+	"github.com/eclipse/paho.golang/packets"
+)
+
+var (
+	ErrPacketNotFound = errors.New("packet not found")
+)
+
+type acksTracker struct {
+	mx    sync.Mutex
+	order []packet
+}
+
+func (t *acksTracker) add(pb *packets.Publish) {
+	t.mx.Lock()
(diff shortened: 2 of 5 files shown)


fracasula:
@alsm I just wanted to give you a quick update. We've been running this code in production for a few days now.
The only issues that we came across was about a `panic` unrelated to these changes. I've fixed it here as well but I'm planning to remove the commit from here and rebase once we manage to merge this [hotfix](https://github.com/eclipse/paho.golang/pull/58).

I've created a Grafana dashboard to monitor these new deployments as well, specifically to see if there would be missing acknowledgments. Given that our VerneMQ is set to drop messages once a specific threshold of unacked messages is reached, looking at these kind of metrics help us detect anomalies.

Screenshot from 2021-06-16 15-14-39

We couldn't see any dropped messages and the `PUBACK`s received by VerneMQ seems to be consistent with what the previous version was giving us.

As a sidenote, I got the [confirmation](https://github.com/hivemq/hivemq-mqtt-client/issues/481#issuecomment-861379255) from the HiveMQ guys that they buffer acknowledgments similarly to what I'm doing here. I think this should give us enough confidence given that they had the client behaving like that for a long time now.

alsm:
I like what this does, I've just been wondering if there's a nice way to do it to make it more customisable. For example Pinger, Auther, Persistence and Router are all interfaces, while the code you've proposed meets your particular need do I/we want that to be the only method available. If I can't come up with something shortly I may merge this as is with a view to coming back to it again in the future.

Issue #59, opened 2021-06-22
Typical example


mlmarius:
I believe it would be very beneficial to have a concise example showing the following:

- set up a connection with and without TLS
- set up automatic reconnect and handle any cleanup necessary when disconnect/reconnect happens
- set up some simple routing

MattBrittan:
@mlmarius this prompted me to raise PR #60 (wrote this some time ago but had not got around to raising a PR). Until this gets committed (assuming its accepted!) you can [view it here](https://github.com/ChIoT-Tech/paho.golang) - see tha `autopaho` folder.

This PR adds 'autopaho'; a wrapper around the `paho` package providing an easy way to access basic functionality (connect, automatically reconnect if connection lost etc) and includes an example publisher/subscriber anong with the docker config files to get it all running. The aim was to provide an easy way to get started with the library as well as example code for those with more complex requirements.

PR #60, opened 2021-06-22
Autopaho - an easy to use wrapper around paho (manages the connection/reconnection process)

MattBrittan:
This PR introduces `autopaho` which wraps the `paho` package with a simple interface that manages the connection. This provides a simple way for new users to utilise the library and also demonstrates its usage. A docker based example is included.

I have not raised an issue re this because I feel its related to #59 and discussed with @alsm a while back.

Apologies for the earlier commit/reverse (was attempting to refresh my fork via the github UI and ended up pushing the changes instead).

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,9 +8,22 @@ Installation and Build
 
 This client is designed to work with the standard Go tools, so installation is as easy as:
 
-```golang
+```bash
 go get github.com/eclipse/paho.golang
 ```
+
+Folder Structure
+----------------
+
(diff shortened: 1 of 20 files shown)


alsm:
Really useful, thanks

samba:
This is great!  @alsm do you know if there are any plans to cut a new tag/release soon?  I'd love to use this in a current prototype, but that would be helped with a proper tag (for go mod).

Issue #61, opened 2021-06-25
Websocket examples


superchalupa:
It's unclear how I would connect to a MQTT server over a websocket.  The older paho client had built in websocket support, but I understand reading through other issues that the design philosophy here is not to do the connection/protocol support in the library.

I looked at gorilla websocket library, but the interface provided by their lib does not match the expected net.Conn interface expected by this paho library and it's not clear to me the best way to achieve a websocket connection. It would be very helpful if we had an example client websocket connection. I'm not tied to gorilla, personally, but that seems like a popular solution. If you have pointers to another library that works out of the box I could probably figure this out, but for future users, an example would likely be helpful.

alsm:
Yeah, something to look at, in the meantime you could just take everything in the old client's websocket.go https://github.com/eclipse/paho.mqtt.golang/blob/master/websocket.go that sets up a websocket connection with gorilla then wraps it in something that provides a net.Conn

superchalupa:
Nice! Thanks for the pointer, that helps immensely.

MattBrittan:
I considered including websocket support in easypaho but doing so would introduce another dependency...thoughts @alsm ?

alsm:
I think it makes sense to add it there, people are going to want websocket connections and it handles creating other types, I don't mind the dependency as autopaho is a separate package so anything we do there doesn't restrict the main client.

MattBrittan:
PR raised. Note that as this is currently a single module the dependency will be loaded regardless (I believe this will be resolved by the "Lazy module loading" coming [in Go 1.17](https://tip.golang.org/doc/go1.17)).

superchalupa:
Thanks, I think that since AutoPaho seems like the correct layered approach to this, I'll close out this issue since websocket support merged into AutoPaho. I'll have to update my stuff to test that out, and will add any further discussion in the AutoPaho repository.

Thanks for the timely and excellent responses!

PR #62, opened 2021-06-27
Add websocket support to AutoPaho

MattBrittan:
Add support for websocket connections to AutoPaho. The docker example has been updated to make use of this.

Ref Issue: #61


alsm:
Super, thanks for adding that

PR #69, opened 2021-07-08
autopaho: set connect packet parameters

### LINKED ISSUE #68: autopaho: pass connect credentials, etc?
Any plans to allow passing of other Connect parameters, such as username and password?

https://github.com/eclipse/paho.golang/blob/a215349036a5852a292890e4fea23585aa19e747/autopaho/net.go#L49-L53

Seems like several of the Connect packet properties would be beneficial to configure: 

https://github.com/eclipse/paho.golang/blob/a215349036a5852a292890e4fea23585aa19e747/paho/cp_connect.go#L7-L17

Should most of those properties be promoted up to the autopaho.ClientConfig and then copied down to each connection?

Alternatively, would it be more suitable to let the autopaho.ClientConfig include a callback method, responsible for generating a *paho.Connect instance?

(I'd be happy to contribute if needed; will be signing ECA shortly.)

samba:
Closes #68 allowing configuration of Connect packet parameters in autopaho.

Slight variations on the proposed interface, but should be essentially the same. 
As this is my first PR on Paho repos, I'm happy to iterate with any feedback provided.

**Testing**

As the autopaho module did not previously have dedicated tests, I have implemented a unittest only for the ClientConfig construction, and related validation of the Connect packet my changes introduced.

I'm also using my fork in an implementation, and it's demonstrably working, correctly authenticating with a broker, etc. 🙂

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,9 @@
-.PHONY: test
-test:
+.PHONY: test unittest
+
+unittest:
+	go test -race -tags=unittest ./autopaho/ -v -count 1
+
+test: unittest
 	go test -race ./packets/ -v -count 1
 	go test -race ./paho/ -v -count 1
+	
--- a/autopaho/auto.go
+++ b/autopaho/auto.go
@@ -81,11 +81,14 @@ type ConnectionManager struct {
 	done chan struct{} // Channel that will be closed when the process has cleanly shutdown
 }
 
+// ResetUsernamePassword clears any configured username and password on the client configuration
 func (cfg *ClientConfig) ResetUsernamePassword() {
 	cfg.connectPassword = []byte{}
 	cfg.connectUsername = ""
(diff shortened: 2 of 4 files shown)


samba:
@MattBrittan  I'd welcome feedback on this, if there are any outstanding concerns. If there are none, what are the next steps to getting this merged?

MattBrittan:
@samba apologies - @alsm has been the only committer to this repo in the past so I had been leaving this for him. Your commit looks good (thanks!); I'll check with @alsm to see how he wants to proceed (it may be that I look after changes to `autoPaho` but did not want to assume that).

alsm:
Thanks, I did look at it and forgot to ask @MattBrittan to look over it as he's more familiar with the autopaho part, as such I'm happy to merge this.

Issue #89, opened 2022-06-14
Does this library support v3.1.1 protocol?


cmackenzie1:
I see some references in the code to protocol version 3, but no way to specify it.

https://github.com/eclipse/paho.golang/blob/master/paho/client.go#L20-L21
https://github.com/eclipse/paho.golang/blob/master/paho/client.go#L211-L212

Could the README.md be updated to reflect the supported protocols?

alsm:
I does not and there are no immediate plans to support v3.1.1. in this library.

alsm:
I think given the edge case disparities between 3.1.1 and 5 support for the former is specifically not a feature that will be supported.

Issue #6, opened 2018-11-02
V3 client in the same style


alsm:
A v3 client in the style of the new v5 client should prove to be a more maintainable (and potentially more correct) implementation than the current v3 client at https://github.com/eclipse/paho.mqtt.golang/
Thought needs to be given to the design/integration; it is possible to provide a single client library support both v3 and v5? If possible is it desirable (would it make the code too messy)?

These questions need to be answered before a 1.0 release, thoughts and ideas welcome.

juliandroid:
You could always rewrite v3 by using v5 library (i.e. v3 to be wrapper around v5), so if you fix something in v5 it will be automatically available in v3.

jjm011:
Do you have a rough estimate for V5 support 1.0 release? What is the current status of this release? I am new to V5 but my project requires a golang mqtt client that uses the V5 standard, so I may be able to help if you can list me a few missing points that need to be addressed for the 1.0 stable release.

alsm:
@juliandroid This client is not going to be API compatible with the previous client, and any v3 client here will be the same. There is some commonality between the two that could be shared, I just need to see how much that is and how easily I can share those parts.

alsm:
@jjm011 I don't really, a 1.0 means the API is fixed, I won't be doing that until I've had more feedback from people using it as right now it's pretty much just me using it. If you have time to do some prototyping with the client and provide feedback on it that would be really helpful.

juliandroid:
@alsm What I meant is create a compatible V3 implementation that wraps your new V5.
So, your V3 implementation "import V5" and every method from the V3 relies(calls) on respective V5 method. Of course since V3 and V5 API is not 1:1 you'll still need to add some code to V3, for example dummy "Token" implementation that returns immediately (assuming V5 is synchronous API).
On the plus side you'll have easy to back port features form V5 to V3 that are not required by the V3 standard (like $share). Those could be provided as additional "extended" interface. Also fixing V5 bugs automatically would be fixed in V3.

Personally, I have written a simplified V3 client that maps current V3 paho implementation because I wanted my application to be (integration) testable, which wasn't possible with current V3 since you need a real mqtt broker.

On that matter I would like to request (V5) API that is fully mockable, so I can write application that doesn't need real V5 client and (real) broker. With the V3 I had to (re)implement Client+Callbacks+Message+Token interfaces to be fully independent.

alsm:
[...]

PR #104, opened 2022-11-10
Initial commit for MQTT v3.1.1 support

viscropst:
A solution for issue #89 ,#6. aid to using paho.golang for MQTT V3.1.1

### DIFF
--- a/packets/auth.go
+++ b/packets/auth.go
@@ -36,6 +36,9 @@ func (a *Auth) String() string {
 
 // Unpack is the implementation of the interface required function for a packet
 func (a *Auth) Unpack(r *bytes.Buffer) error {
+	if isVer4() {
+		return nil
+	}
 	var err error
 
 	success := r.Len() == 0
@@ -70,6 +73,9 @@ func (a *Auth) Buffers() net.Buffers {
 
 // WriteTo is the implementation of the interface required function for a packet
 func (a *Auth) WriteTo(w io.Writer) (int64, error) {
+	if isVer4() {
+		return 0, nil
+	}
 	cp := &ControlPacket{FixedHeader: FixedHeader{Type: AUTH}}
 	cp.Content = a
 
--- a/packets/compat_spec.go
+++ b/packets/compat_spec.go
@@ -1,0 +1,140 @@
+package packets
+
+import (
+	"bytes"
+)
+
+const minVer = 4
+const maxVer = 5
+
+var _protocolVersion byte = maxVer
+
+// protocolVersion for swapping or getting MQTT version in packet
+func protocolVersion(vers ...byte) byte {
+	var ver byte
+	if len(vers) > 0 {
+		ver = vers[0]
+	}
+	if minVer <= ver && ver <= maxVer {
+		_protocolVersion = ver
+	}
+	return _protocolVersion
+}
+
+func isVer4() bool {
+	return _protocolVersion == 4
+}
+
+type propPack struct {
+	typ byte
+}
+
+func (p propPack) Pack(prop *Properties, cp *bytes.Buffer) {
+	if isVer4() {
+		return
+	}
+	idvp := prop.Pack(p.typ)
+	encodeVBIdirect(len(idvp), cp)
+	cp.Write(idvp)
+}
+
+func (p propPack) Unpack(r *bytes.Buffer, prop *Properties) error {
+	if isVer4() {
+		return nil
+	}
+	err := prop.Unpack(r, p.typ)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func genPropPack(p byte) propPack {
+	return propPack{typ: p}
+}
+
+func connackReasonFromV311(code byte) byte {
+	switch code {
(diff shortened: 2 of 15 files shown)


ewirch:
@viscropst, why don't you use github.com/eclipse/paho.mqtt.golang for v3 MQTT?

mvrhov:
Because the interface of that library is horrible and it looks like it was done by someone who has never seen go before.

MattBrittan:
I agree that the interface of the V3.1.1 library is not as good (this library was started afresh partly because of the issues with the V3 client API). However the V3 client is a more complete implementation of the protocol (V5 client does not yet support persistence) so I'd still recommend it for most users (if you want to use the V5 client then it's important to understand its limitations). The V3 client is now stable (I've had instances running constantly for over a year on unstable links) and the reality is that most programs will probably only reference the library on 5-10 lines so, the sub-optimal API is probably not a massive issue. 

Ref this specific issue; I had been waiting for @alsm to comment here, but as it's been a few months (and there has been some reaction) I'll add my thoughts. As you noted a request for V3 functionality was made some time ago (#6 - [V3 client in the same style](https://github.com/eclipse/paho.golang/issues/6)) and this was closed by @alsm with:

>What I'm saying is that superficially it might seem like only a few changes to make this client support 3.1.1 but it will actually be a lot more work to make a client that is fully conformant to both specs. I don't think it can be reasonably done in the same codebase. If @viscropst wants to try I'm more than happy to be proven wrong 😁 and if it works will happily accept it, but what I don't want is to make the current v5 client harder or more awkward to use.

My biggest concerns are the complexity of use and the complexity of further development (the lesson learnt from the V3 client is that seemingly simple options can be a lot of work to maintain and make future enhancements much more difficult). As such I'm keen for the V5 client to remain as simple as possible (whilst fully supporting MQTT V5).

I personally feel that implementing persistence should be the highest priority for this project (because currently the client does not fully implement the MQTT spec as it stands) and am concerned that implementing V3 will make this more difficult (there are some subtle differences in the specs). This will be a fairly big change and I have put some time looking into approaches (but unfortunately something else always came up before I was able to finish an implementation).

If a V3 client with this API is desired then I'd prefer it to be implemented as a sub project that is implemented in the same style as the V5 client (so you can just swap the import path to get the V3 client) - perhaps with a package providing a connect function that supports both. This would lead to more duplication of code but would remove the need for all of the `if !isVer4()` and avoid limiting the V5 client. Should both clients be implemented together (as per this PR) then I think we would need to accept that this is a V5 first client (and V3 is a bit of a second class citizen because guidance from the V5 spec will be followed by default).

I have not looked at the implementation in detail (sorry) but did note that it appears to be using a global variable (`packets/_protocolVersion`) which I'm not keen on (will work until someone tries to bridge MQTTV3 and V5 brokers with the client and looks like it's technically a race condition when there are two broker connections). I'll see what further feedback this gets (and whether @alsm responds) before considering looking further (Al created this client and I just added the AutoPaho part so had only been looking at PR's in that area).

@viscropst - thanks very much for your work on this and I do hope that we can use it; my caution is mainly due to the many hours I've spent trying to untangle the V3.1.1 client (and once you add a feature it's very hard to remove it!).

mvrhov:
I haven't looked deeply into the library. I've just tested both libs v3 and v5. But IMHO you should provide only interface and if one really needs persistence, than they should implement it themselves.
There is nothing that prevent this library then to provide a persistence as a different package.

MattBrittan:
>  if one really needs persistence, than they should implement it themselves

Sorry - the terminology I used may have been confusing (I have edited my response to clarify). By persistence I mean storing the session state (see [4.1.1 in the spec](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901231)). MQTT supports various QOS levels enabling you to guarantee delivery/receipt; this client does not currently support this feature (it will allow you to specify QOS1/2 but will never resend a message and will just respond yes when asked for confirmation if it has received a message). See issue #25 for more info.

For some use-cases (i.e. when the latest value is all you are interested in) this is fine, but if missing messages is a problem, then you cannot currently rely on this client (it's why I have not moved to it yet).

>  if one really needs persistence, than they should implement it themselves. There is nothing that prevent this library then to provide a persistence as a different package.

Unfortunately it's fairly complex to implement and tied deeply into the protocol. The library would need major modifications to support a third party persistence package and such code would need to be pretty tightly coupled (i.e. when receiving an acknowledgment the library needs to know about the message the acknowledgment refers to so it can respond correctly). This is not something that a normal user could, or should be expected to, implement (I have had a few goes; the challenge is implementing it in a way that does not add too much complexity).

alsm:
I appreciate the effort you've put in to doing this but while this offer superficial MQTT 3.1.1 support there are differences in behaviour that will not be reflected and make it not a replacement for other 3.1.1 clients. As such I think 3.1.1 support is specifically not going to be a feature of this library.

alsm:
Closing as 3.1.1 support is no longer something I plan to implement

PR #107, opened 2022-12-04
Add RPC Timout request and prevent blocking while waiting response

### LINKED ISSUE #108: Add RPC Request timeout and prevent request blocking
this issue is related to this PR https://github.com/eclipse/paho.golang/pull/98

> The most significant issue was the rpc.Handler.Request function, in which the go routine could block indefinitely waiting for a response from the server.

and I found that there are missed fixes from the last PR on the following line https://github.com/eclipse/paho.golang/blob/e818d6af9263c1ff1c68749d9e87defb8f51690f/paho/extensions/rpc/rpc.go#L76

* add context to the Request channel to prevent blocking
* add request timeout

yogihardi:
closes #108

--

this PR is related to this request https://github.com/eclipse/paho.golang/pull/98

> The most significant issue was the rpc.Handler.Request function, in which the go routine could block indefinitely waiting for a response from the server.

I found that there are missed fixes from the last PR on the following line
https://github.com/eclipse/paho.golang/blob/e818d6af9263c1ff1c68749d9e87defb8f51690f/paho/extensions/rpc/rpc.go#L76

also, I proposed request timeout on RPC Request

### DIFF
--- a/paho/extensions/rpc/rpc.go
+++ b/paho/extensions/rpc/rpc.go
@@ -2,12 +2,15 @@ package rpc
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
 
 	"github.com/eclipse/paho.golang/paho"
 )
+
+const defaultRequestTimeout = time.Second * 10
 
 // Handler is the struct providing a request/response functionality for the paho
 // MQTT v5 client
@@ -54,7 +57,7 @@ func (h *Handler) getCorrelIDChan(cID string) chan *paho.Publish {
 	return rChan
 }
 
-func (h *Handler) Request(ctx context.Context, pb *paho.Publish) (*paho.Publish, error) {
+func (h *Handler) Request(ctx context.Context, pb *paho.Publish) (resp *paho.Publish, err error) {
 	cID := fmt.Sprintf("%d", time.Now().UnixNano())
 	rChan := make(chan *paho.Publish)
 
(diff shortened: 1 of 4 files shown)


paho/extensions/rpc/rpc.go
```diff
@@ -68,13 +71,19 @@ func (h *Handler) Request(ctx context.Context, pb *paho.Publish) (*paho.Publish,
...
-	_, err := h.c.Publish(ctx, pb)
+	_, err = h.c.Publish(ctx, pb)
 	if err != nil {
 		return nil, err
 	}
 
-	resp := <-rChan
-	return resp, nil
+	select {
+	case <-time.After(defaultRequestTimeout):
```
MattBrittan, line 80: I would argue that providing a context here should be sufficient (if the user wants a timeout then they can set one, otherwise there is no timeout). Setting 10 seconds seems arbitrary (say the RPC job takes a minute to complete; the user has no way to turn this timeout off).

MattBrittan (commented):
Thanks for your PR - I have added a few comments. Please note that you will need to complete the CLA (see [contributingh.md](https://github.com/eclipse/paho.golang/blob/master/CONTRIBUTING.md)) before this can be accepted (this process ensures that we have verification that all contributions are correctly licensed).

paho/extensions/rpc/rpc.go
```diff
@@ -2,13 +2,16 @@ package rpc
...
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
 
 	"github.com/eclipse/paho.golang/paho"
 )
 
+const defaultRequestTimeout = time.Second * 10
```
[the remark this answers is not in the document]
yogihardi: ok, I've removed it

paho/extensions/rpc/rpc.go
yogihardi: thanks @MattBrittan, I have set the requestTimeout as an optional config

yogihardi:
hi @MattBrittan ,

thank you for your review.
i couldn't verify my eca, even though my email address is same and correct.

I got the following error message:
```
Failures:
Anonymous (<my email account>@****.com)
```

MattBrittan:
> thank you for your review. i couldn't verify my eca, even though my email address is same and correct.

I have manually checked and the ECA system does not appear to have your address on file (note that email addresses used in Github are publicly available and easily retrieved from [the patch](https://patch-diff.githubusercontent.com/raw/eclipse/paho.golang/pull/107.patch)). Can you please confirm that you have [created an account](https://accounts.eclipse.org/user/login?destination=user/eca) and signed the ECA?

I'm OK with the PR itself (but unfortunately cannot accept it until we get the ECA sorted). Personally I'd probably just return the `errCtx` because most users would expect to check for the standard context errors but can also see some advantages to your approach.

PR #112, opened 2023-01-10
Add RPC Timout request and prevent blocking while waiting response

### LINKED ISSUE #108: Add RPC Request timeout and prevent request blocking

yogihardi:
closes #108

--

this PR is related to this request https://github.com/eclipse/paho.golang/pull/98

> The most significant issue was the rpc.Handler.Request function, in which the go routine could block indefinitely waiting for a response from the server.

I found that there are missed fixes from the last PR on the following line
https://github.com/eclipse/paho.golang/blob/e818d6af9263c1ff1c68749d9e87defb8f51690f/paho/extensions/rpc/rpc.go#L76

also, I proposed request timeout on RPC Request

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 bin/
 autopaho/cmd/rpc/.env
+.idea
--- a/paho/cmd/rpc/main.go
+++ b/paho/cmd/rpc/main.go
@@ -179,7 +179,10 @@ func main() {
 
 	fmt.Printf("Connected to %s\n", *server)
 
-	h, err := rpc.NewHandler(ctx, c)
+	h, err := rpc.NewHandler(ctx, rpc.HandlerInput{
+		Client:         c,
+		RequestTimeout: nil, // no timeouts
+	})
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/paho/extensions/rpc/rpc.go
+++ b/paho/extensions/rpc/rpc.go
@@ -13,21 +13,29 @@ import (
 // MQTT v5 client
 type Handler struct {
 	sync.Mutex
-	c          *paho.Client
-	correlData map[string]chan *paho.Publish
+	c              *paho.Client
+	correlData     map[string]chan *paho.Publish
+	requestTimeout *time.Duration
 }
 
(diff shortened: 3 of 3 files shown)


yogihardi:
hi @MattBrittan ,

I recreate the PR since unable to update the author's email on each commits and previous PR

MattBrittan:
Apologies for not responding to this sooner (I had not previously been looking into PR's on the `paho.golang/paho` side of things). f you are able to update this so it can be merged with the current code base then I can have a look. 

Note that I'm not sure if adding `RequestTimeout` is worthwhile; `Request` is already passed a `Context` so I would have thought that would be sufficient (i.e. just need to add the select around `resp := <-rChan`.

nych:
> Apologies for not responding to this sooner (I had not previously been looking into PR's on the `paho.golang/paho` side of things). f you are able to update this so it can be merged with the current code base then I can have a look.
> 
> Note that I'm not sure if adding `RequestTimeout` is worthwhile; `Request` is already passed a `Context` so I would have thought that would be sufficient (i.e. just need to add the select around `resp := <-rChan`.

Introducing the `RequestTimeout` member could be practical since it eliminates the need to explicitly specify a timeout for each request. However, this approach also imposes a limitation by preventing the definition of unique timeouts for individual requests when a global timeout is set. Personally, I find the more flexible solution proposed by @MattBrittan to be preferable.

MattBrittan:
I have raised PR #620 with the alternate approach, are you happy with that?

Note: The RPC stuff could really do with a rewrite (it's become a bit of a mess).

MattBrittan:
Closing this one as I believe my alternate approach resolves the issue. Would appreciate any further work on the RPC code (it's a mess).

Issue #115, opened 2023-01-18
Autopaho CleanStart forced to true always


ashtonian:
**Describe the bug**
autopaho is currently forcing clean start. 

**To reproduce**
try to connect to autopaho without a clean start. 

**Expected behaviour**
To be able to use autopaho with a clean start.

Ashtonian:
Not sure if I need to configure the in memory stuff as well in order for this to work. Any thoughts?

MattBrittan:
Please see my comment on #113 (these two issues should probably be merged; ideally all being rolled into #25). My main issue with this is that without support for persistence setting `CleanSession` to false means we are not complying with the MQTT spec which details what is included in the session (see [here](https://github.com/eclipse/paho.golang/blob/d63b3b28d25ff73076c8846c92c4d062503e646e/paho/client.go#L458) re handling ACK packets). This would be a real issue at QOS2 (because we could keep receiving, for example, `PUBREL`, which we ignore, so from the broker perspective the message workflow never completes).

As such I think this would need a warning comment and my preference would be that it continues to default to `CleanStart=true` (I realise doing it the way you have is easier but this is a breaking change and even through we are pre 1.0 I'd prefer to avoid those if possible, especially one like this that users are likely to miss). This would not be such an issue if persistence was implemented; implementing that is likely to result in other breaking changes (and is not a small task).

akindestam:
+1 for this, since connecting with `CleanStart=false` would be useful for my use case even without the full persistence support (my use case is mainly QoS 1)

I do agree that it might need to default to CleanStart=true somehow, though...

MattBrittan:
I'm going to close this as a duplicate of #25 (because support for a session is needed before CleanStart can be enabled). Support has been implemented in my test repo (see link in that issue) and I am currently testing it. Assistance with testing would be welcomed :-).

PR #116, opened 2023-01-18
allow clean start with auto paho

ashtonian:
closes https://github.com/eclipse/paho.golang/issues/115 

Ensure you have a valid, signed [ECA](https://www.eclipse.org/legal/ECA.php) for the email address associated with your Github account (or you sign off your PR with an ECA'd address) 👍🏽


MattBrittan:
Thanks for your contribution but I'm going to close this because it will be dealt with in the upcoming PR that sorts out QOS1/2 (did not want to accept this as it would make the following change a breaking one).

The question is how should reconnects be handled; the approach I'm taking is that if the user sets `CleanSession` to true in autopaho then the first connection will be clean whereas following ones will not (the idea being that reconnects should be transparent to the user so setting `CleanSession` to false on those makes sense).

PR #117, opened 2023-01-21
Fix the issue c.expectConnack goroutine doesn't exit #101

### LINKED ISSUE #101: expectConnack goroutine not exit
image

if  connCtx  timeout  
expectConnack not quit

fix bug
var (
		caPacket    *packets.Connack
		caPacketCh  = make(chan *packets.Connack)
		caPacketErr = make(chan error,1)
	)

tomatod:
This patch is for fixing #101. As I commented in the GitHub issue, it can be easily reproduced and causes goroutine leak (memory leak). After the patch is applied, it cannot be reproduced by the reproduction code in the GitHub issue.

To prevent the issue c.expectConnack goroutine doesn't exit, caPacketCh and caPacketErr is changed to buffered channels from unbuffered one. comment are also added so that the buffers won't get deleted in future fixes.

closes #101

### DIFF
--- a/paho/client.go
+++ b/paho/client.go
@@ -220,8 +220,12 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
 	c.debug.Println("waiting for CONNACK/AUTH")
 	var (
 		caPacket    *packets.Connack
-		caPacketCh  = make(chan *packets.Connack)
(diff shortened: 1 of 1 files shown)


MattBrittan:
Thanks for doing this - sorry for the delay in reviewing.

Issue #132, opened 2023-06-01
Auth packet's return code is defined, but not used.


tomatod:
**Describe the bug**
Auth packet's return code is difined, but not used.

define:
https://github.com/eclipse/paho.golang/blob/master/packets/auth.go#L17-L22

not used (value is hard corded):
https://github.com/eclipse/paho.golang/blob/master/paho/client.go#L429-L444

I think that there may be still other similar implemention (not only Auth).

**Expected behaviour**
Each hard coded values are replaced by each defined constant.

PR #133, opened 2023-06-01
Hard coded return codes of Auth packet are replaced by each defined constant

tomatod:
I think it's better to replace hard-coded return code by constant, because there are already defined constants. It's a little confusing.

#132


alsm:
Thanks for your PR

PR #136, opened 2023-06-06
Fix not using max MID (full band of MID)

### LINKED ISSUE #135: MIDs doesn't use full band of MID
**Describe the bug**
Request() of MIDs issues MID only from 1 to 65534. 65535 isn't issued. I think it seems to be unexpected behavior.

**To reproduce**
The following test can reproduce the unexpected behavior in paho/message_ids_test.go.
```
func TestUseFullBandOfMID(t *testing.T){
        m := &MIDs{index: make([]*CPContext, midMax)}
        cp := &CPContext{}
        for i := uint16(0); i < midMax; i++ {
                v, _ := m.Request(cp)
                t.Log("result: ", v,  "expect:", i+1)
                assert.Equal(t, i+1, v)
        }
}
```

**Expected behaviour**
Request() of MIDs issues MID from 1 to 65535. According to spec, it seems to be not problem to set 65535 to "Packet Identifier".

https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901026
> The Variable Header component of many of the MQTT Control Packet types includes a Two Byte Integer Packet Identifier field. These MQTT Control Packets are PUBLISH (where QoS > 0), PUBACK, PUBREC, PUBREL, PUBCOMP, SUBSCRIBE, SUBACK, UNSUBSCRIBE, UNSUBACK.

https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901008
> Two Byte Integer data values are 16-bit unsigned integers in big-endian order: the high order byte precedes the lower order byte. This means that a 16-bit word is presented on the network as Most Significant Byte (MSB), followed by Least Significant Byte (LSB).

**Additional context**
In #134, the implemention of MIDs is discussed. So when I was reading the code, I noticed this issue. It's a little confusing, so I think it's better to fix it if it's not expected behavior.

tomatod:
This fix correspond to #135. 

Note that In this implemention, whether last MID is blank is also checked. (In current implemention of "master", checking  up to last MID - 1)

close #135

### DIFF
--- a/paho/message_ids.go
+++ b/paho/message_ids.go
@@ -48,7 +48,7 @@ type CPContext struct {
 type MIDs struct {
 	sync.Mutex
 	lastMid uint16
-	index   []*CPContext
+	index   []*CPContext // index of slice is (messageid - 1)
 }
 
 // Request is the library provided MIDService's implementation of
@@ -56,20 +56,26 @@ type MIDs struct {
 func (m *MIDs) Request(c *CPContext) (uint16, error) {
 	m.Lock()
 	defer m.Unlock()
-	for i := uint16(0); i < midMax; i++ {
-		v := (m.lastMid + i) % midMax + 1
-		if m.index[v-1] != nil {
-			// In some case, additionally check if last MID is blank at last loop, because it may be already blank here.
-			// When last MID is 0, MIDs was just initialized and checking isn't required.
-			if !(m.lastMid != 0 && i == midMax - 1 && v != m.lastMid  && m.index[m.lastMid-1] == nil) {
-				continue
-			}
-			v++
+
+	// Scan from lastMid to end of range.
+	for i := m.lastMid; i < midMax; i++ {
+		if m.index[i] != nil {
+			continue
 		}
-		m.index[v-1] = c
-		m.lastMid = v
-		return v, nil
+		m.index[i] = c
+		m.lastMid = i + 1
+		return i + 1, nil
 	}
+	// Scan from start of range to lastMid
+	for i := uint16(0); i < m.lastMid; i++ {
+		if m.index[i] != nil {
+			continue
+		}
+		m.index[i] = c
+		m.lastMid = i + 1
+		return i + 1, nil
+	}
+
 	return 0, ErrorMidsExhausted
 }
 
--- a/paho/message_ids.go
+++ b/paho/message_ids.go
@@ -56,16 +56,19 @@ type MIDs struct {
 func (m *MIDs) Request(c *CPContext) (uint16, error) {
 	m.Lock()
 	defer m.Unlock()
-	for i := uint16(1); i < midMax; i++ {
-		v := (m.lastMid + i) % midMax
-		if v == 0 {
-			continue
+	for i := uint16(0); i < midMax; i++ {
+		v := (m.lastMid + i) % midMax + 1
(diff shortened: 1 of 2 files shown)


paho/message_ids.go
```diff
@@ -56,16 +56,19 @@ type MIDs struct {
...
-	for i := uint16(1); i < midMax; i++ {
-		v := (m.lastMid + i) % midMax
-		if v == 0 {
-			continue
-		}
-		if inuse := m.index[v]; inuse == nil {
-			m.index[v] = c
-			m.lastMid = v
-			return v, nil
+	for i := uint16(0); i < midMax; i++ {
```
MattBrittan, line 59: Please don't feel obliged to accept this suggestion. I'm mainly adding it because I found the current version a little hard to understand (some of that is the comment, some relates to uint16 overflows...).

Might I suggest:

```
	for i := uint16(0); i < midMax; i++ {
		v := (m.lastMid+i)%midMax + 1
		if m.index[v-1] != nil {
			continue
		}
		m.index[v-1] = c
		m.lastMid = v
		return v, nil
	}
	// It may be possible to reuse lastMid
	if m.lastMid != 0 && m.index[m.lastMid-1] == nil {
		m.index[m.lastMid-1] = c
		return m.lastMid, nil
	}

	return 0, ErrorMidsExhausted
```

or, even better in my opinion (because it's very obvious what it's doing and avoids the need to take a remainder):

```
func (m *MIDs) Request(c *CPContext) (uint16, error) {
	m.Lock()
	defer m.Unlock()
	for i := m.lastMid + 1; i != 0; i++ { // scan from lastMid + 1 to end of range
		if m.index[i-1] != nil {
			continue
		}
		m.index[i-1] = c
		m.lastMid = i
		return i, nil
	}
	for i := uint16(0); i < m.lastMid; i++ { // scan from start of range to lastMid (inclusive)
		if m.index[i] != nil {
			continue
		}
		m.index[i] = c
		m.lastMid = i + 1
		return i + 1, nil
	}

	return 0, ErrorMidsExhausted
}
```

paho/message_ids.go
```diff
@@ -75,14 +78,14 @@ func (m *MIDs) Request(c *CPContext) (uint16, error) {
 func (m *MIDs) Get(i uint16) *CPContext {
 	m.Lock()
 	defer m.Unlock()
-	return m.index[i]
+	return m.index[i-1]
```
MattBrittan, line 81: It may be worth adding a comment in the `struct` definition ```index []*CPContext // index is messageid-1``` (just to avoid future misunderstandings)

MattBrittan (commented):
Took a quick look - works well but I've added a few subjective comments :-)

paho/message_ids.go
tomatod: Thanks for your review !! Your second opinion is easy to read, so I reflected it to my implemention with a few change.

paho/message_ids.go
tomatod: Thanks you :)
Current original (master) inplemention is index == messageid, so I also think the comment is worth.

alsm:
Thanks for this, I liked the original that it was one loop to search all the mids and find the next one but this is probably easier to read

PR #139, opened 2023-06-09
implement simplified pinger and add pinger tests

### LINKED ISSUE #137: PingHandler will not time time out
**Describe the bug**
The logic of `PingHandler` means that there will never be a timeout, and `pingFailHandler` will never be called. I'm working on getting out a PR for this. 

**To reproduce**
Wrote a test which should pass if `PingHandler` times out, but the test fails:
```go
package paho

import (
	"net"
	"testing"
	"time"

	"github.com/stretchr/testify/assert"
)

func TestPingerPingHandlerTimeout(t *testing.T) {
	fakeServerConn, fakeClientConn := net.Pipe()
	defer fakeServerConn.Close()
	defer fakeClientConn.Close()

	serverConnBuffer := make([]byte, 1024)

	go func() {
		// keep reading from fakeServerConn and discard the values. exit when fakeServerConn is closed.
		for {
			_, err := fakeServerConn.Read(serverConnBuffer)
			if err != nil {
				return
			}
		}
	}()

	pfhValues := make(chan error)
	pfh := func(e error) {
		pfhValues <- e
	}

	pingHandler := DefaultPingerWithCustomFailHandler(pfh)

	go func() {
		pingHandler.Start(fakeClientConn, time.Second)
	}()
	defer pingHandler.Stop()

	select {
	case <-time.After(time.Minute):
		t.Error("pingHandler did not timeout after a minute of waiting")
	case err := <-pfhValues:
		assert.EqualError(t, err, "ping resp timed out")
	}
}
``` 

**Expected behaviour**
`PingHandler` should eventually call `pingFailHandler` if `PingResp()` is never called.

**Additional context**
Looks like this happens because `p.lastPing` is reset every time a `PINGREQ` is sent (approximately every `pt` duration), so `time.Since(p.lastPing)` doesn't get the chance to get to 1.5x `pt` for a timeout to be detected.

vishnureddy17:
close #137 

Here's a proposal for a pinger design. Essentially, it sends a PINGREQ and checks for a PINGRESP in lockstep on a keepalive time interval. It is simpler than the currently existing one, but still meets the requirements for the MQTT spec. IMO, keeping it simple is better, especially since advanced users can provide their own implementation if they want to. 

I also added some tests, though the tests do add a non-trivial amount of runtime to the test suite.

### DIFF
--- a/paho/client.go
+++ b/paho/client.go
@@ -219,7 +219,7 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
 
 	c.debug.Println("waiting for CONNACK/AUTH")
 	var (
-		caPacket    *packets.Connack
+		caPacket *packets.Connack
 		// We use buffered channels to prevent goroutine leak. The Details are below.
 		// - c.expectConnack waits to send data to caPacketCh or caPacketErr.
 		// - If connCtx is cancelled (done) before c.expectConnack finishes to send data to either "unbuffered" channel,
--- a/paho/pinger.go
+++ b/paho/pinger.go
@@ -4,27 +4,27 @@ import (
 	"fmt"
 	"net"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/eclipse/paho.golang/packets"
 )
 
 // PingFailHandler is a type for the function that is invoked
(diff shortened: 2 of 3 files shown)


vishnureddy17:
Closing this as it is out of date.

Issue #138, opened 2023-06-09
Improve logging for tests


vishnureddy17:
Currently, running the tests results in a bunch of logs being dumped out, making it hard interpret the results test results. It'd be good to at least have the option to suppress them.

MattBrittan:
I believe that #172 addresses this; logging is now via the `test` package and this means that logs are only output when a test fails (which I think will be appropriate in most cases).

vishnureddy17:
Yep, going to close this.

PR #140, opened 2023-06-13
Only print out full logging in testing when `-args -fullTestLogs` is appended to `go test`

alsm:
`-args -fullTestLogs` is appended to `go test`

#138

### DIFF
--- a/paho/client_test.go
+++ b/paho/client_test.go
@@ -3,6 +3,7 @@ package paho
 import (
 	"context"
 	"errors"
+	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


MattBrittan:
Note that I've taken a different approach in the big PR I'm working on. I'm using the `Log`/`Logf` functions provided by `testing.Test` (see [here](https://github.com/ChIoT-Tech/paho.golang/blob/store/paho/log/test.go#L25)). The benefit of this is that the logs are only output if the test fails (which is likely to be the only time you are really interested in them).

Issue #134, opened 2023-06-01
Thoughts on how to move the project forward


vishnureddy17:
_Continuing the dialogue from #130._

Hi folks, I wanted to open this issue to discuss some thoughts I had on how to move the project forward. Ostensibly, the biggest missing feature is persisting sessions across network connections, which is required to properly support CleanStart=false with QoS > 0 PUBLISHes. However, I see some issues that I think need to be addressed before adding persistent session support. I feel that if these issues get addressed, it will make adding persistence support easier. 

Before I go into discussing these issues, I want to make it clear that this is in no way intended to disparage @alsm's excellent work on this so far. I really appreciate that effort, as well as @MattBrittan's efforts on this project.

# Session state is inherently coupled with packet identifier allocation and tracking
Currently, the packet identifiers are assigned and tracked using the implementation in [message_ids.go](https://github.com/eclipse/paho.golang/blob/42b88ebe91dbe35afe801acc4ad1c0d4a7cbd506/paho/message_ids.go). In order to support persistence, we either need to:
1. Make sure that the persistence implementation being used is always kept in sync with the `MIDService` being used.
2. Add a function to `MIDService` to mark arbitrary packet identifiers as non-free when connecting with an existing session

If we go with option 1, I think it would make sense to combine the functionality of `Persistence` and `MIDService` into a broader session state interface.

# It is difficult to stay spec-compliant with packet timeouts, especially once persistent sessions come into the picture
Currently, the client has a timeout for PUBLISH QoS > 0, SUBSCRIBE, and UNSUBSCRIBE operations in case the expected response packet doesn't come back from the server in a certain amount of time. There is a discussion about it in #23. This is not spec-compliant and will cause issues that will become more acute with persistent sessions in the picture (timeout on CONNECT, however, is fine). Instead, it is the job of the keep-alive mechanism (PINGREQ/PINGRESP) to detect when things have gone wrong with the connection. It looks like the keep-alive mechanism is not wired up to the client and instead just logs an error when a PINGRESP doesn't come back in time. Here's how I believe this should be handled:
- If the client has not sent any outgoing control packets in the past keepalive time interval, send a PINGREQ packet.
- If we do not receive a PINGRESP packet after sending a PINGREQ packet and the keepalive interval has passed, cancel the outstanding PUBLISH QoS > 0, SUBSCRIBE, and UNSUBSCRIBE operations and close the network connection to the server.

[...]

PR #141, opened 2023-06-13
Allow initializing MIDs with in-use IDs, remove Clear(), use RWMutex for Get()

vishnureddy17:
As discussed in #134, we need a way to claim packet IDs as a prerequisite for connecting with existing sessions. This implements that. In addition, I put an RWMutex in since calls to Get() can safely happen in parallel.

### DIFF
--- a/paho/client.go
+++ b/paho/client.go
@@ -138,7 +138,7 @@ func NewClient(conf ClientConfig) *Client {
 		c.Persistence = &noopPersistence{}
 	}
 	if c.MIDs == nil {
-		c.MIDs = &MIDs{index: make([]*CPContext, int(midMax))}
+		c.MIDs = NewMIDs(nil)
 	}
 	if c.PacketTimeout == 0 {
 		c.PacketTimeout = 10 * time.Second
@@ -219,7 +219,7 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
 
 	c.debug.Println("waiting for CONNACK/AUTH")
 	var (
-		caPacket    *packets.Connack
+		caPacket *packets.Connack
 		// We use buffered channels to prevent goroutine leak. The Details are below.
 		// - c.expectConnack waits to send data to caPacketCh or caPacketErr.
 		// - If connCtx is cancelled (done) before c.expectConnack finishes to send data to either "unbuffered" channel,
--- a/paho/message_ids.go
+++ b/paho/message_ids.go
@@ -37,7 +37,6 @@ type MIDService interface {
 	Request(*CPContext) (uint16, error)
 	Get(uint16) *CPContext
 	Free(uint16)
-	Clear()
 }
 
 // CPContext is the struct that is used to return responses to
@@ -56,6 +55,17 @@ type MIDs struct {
 	sync.RWMutex
 	lastMid uint16
 	index   []*CPContext // index of slice is (messageid - 1)
+}
+
+func NewMIDs(inUse map[uint16]*CPContext) *MIDs {
+	m := &MIDs{index: make([]*CPContext, midMax)}
+	for mid, c := range inUse {
+		m.index[mid-1] = c
+		if mid > m.lastMid {
+			m.lastMid = mid
+		}
+	}
+	return m
 }
 
 // Request is the library provided MIDService's implementation of
@@ -86,21 +96,6 @@ func (m *MIDs) Request(c *CPContext) (uint16, error) {
 	return 0, ErrorMidsExhausted
 }
 
-// ClaimID is the library provided MIDService's implementation of
-// the required interface function()
-func (m *MIDs) ClaimID(i uint16, c *CPContext) error {
-	m.Lock()
-	defer m.Unlock()
-	if m.index[i-1] != nil {
-		return ErrorMidInUse
-	}
-	m.index[i-1] = c
-	if i > m.lastMid {
-		m.lastMid = i
-	}
-	return nil
-}
-
 // Get is the library provided MIDService's implementation of
 // the required interface function()
 func (m *MIDs) Get(i uint16) *CPContext {
@@ -115,10 +110,4 @@ func (m *MIDs) Free(i uint16) {
 	m.Lock()
 	m.index[i-1] = nil
 	m.Unlock()
-}
-
-// Clear is the library provided MIDService's implementation of
-// the required interface function()
-func (m *MIDs) Clear() {
-	m.index = make([]*CPContext, int(midMax))
 }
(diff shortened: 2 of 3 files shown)


MattBrittan:
`paho` never reconnects (so, unless this changes, loading `MIDs` would be done in another package, i.e. `AutoPaho`). As such, I wonder if adding `Get()` is the right approach (not criticising your PR, just offering an alternative viewpoint). 

An alternative would be to add a `NewMids(c []*CPContext) *MIDs` (or perhaps `NewMids(map[uint16]*CPContext) *MIDs` might be simpler to use) that can be called to create a pre-populated `MIDs`. This would make it clear that `MIDs` should only be setup before the connection is established and not altered afterwards (personally I'm not sure that `Clear()` is needed; the difference between `Clear` and creating a whole new `MIDs` seems minor).

vishnureddy17:
Good points, I'll find some time to update this PR.

vishnureddy17:
PR updated based on the feedback. 

As a side note, I think that using names like "message ID" and "MID" should be avoided because the MQTT spec using "Packet ID" instead.

vishnureddy17:
@MattBrittan I just realized that this would prevent users from making their own `MIDService`. Marking a Packet ID as "in use" needs to be done with a method in the `MIDService` interface. Thoughts? 

Do we even need to support custom packet ID allocators? @alsm, did you have a use-case for custom packet ID allocators in mind when you implemented this?

MattBrittan:
> Marking a Packet ID as "in use" needs to be done with a method in the MIDService interface

Sorry - I'm not seeing (possibly missing something!) why this is a requirement? 'paho' only needs access to `Request`, `Get` and `Free`.  Someone creating a new implementation would have access to their implementation (i.e. direct access to the `struct`, or additional functions not included in the interface).

>Also, I think claiming Packet IDs should be done right before sending a connect

I guess this is possible in the current design (just load the data the first time one of the functions is called). The alternative would be a callback just before sending `connect` (somewhere around [here](https://github.com/eclipse/paho.golang/blob/master/paho/client.go#L210)); we will need a few more callbacks (or similar) to implement persistent sessions (once the connection is up any unacknowledged messages will need to be resent). However I feel that trying to cater for this now is putting the horse before the cart (we need to work out how session persistence will work and then change `paho` to suit).

If `AutoPaho` was to be modified to support a custom `MIDService` then this may become an issue (because I would expect `AutoPaho` to provide a store which it would need to retrieve upon connection). However I don't really see any need for `AutoPaho` to support a custom implementation of `MIDService` (my aim is for `AutoPaho` to support most basic use-cases; if someone wants to do something unusual then they can use `paho` directly).

vishnureddy17:
> Sorry - I'm not seeing (possibly missing something!) why this is a requirement? 'paho' only needs access to `Request`, `Get` and `Free`. 

Paho also needs access to some way to mark a packet ID as in-use if it is connecting with an existing session, no? That's why I'm thinking that the marking of Packet IDs as in-use needs to be a part of the MIDService interface. Paho will need to get the in-use packet IDs from the persistence implementation, and that may require I/O, so I don't think marking Packet IDs as in-use should be done in the initialization.

> I guess this is possible in the current design (just load the data the first time one of the functions is called). The alternative would be a callback just before sending connect (somew[here](https://github.com/eclipse/paho.golang/blob/master/paho/client.go#L210) around here); we will need a few more callbacks (or similar) to implement persistent sessions (once the connection is up any unacknowledged messages will need to be resent).

Yep, I think around there is sounds right. Before sending the CONNECT packet, Paho will get the relevant session state from the the persistence implementation and mark the relevant packet IDs as in-use. When the CONNACK comes back, Paho will do whatever delivery retry it needs to do per the MQTT requirements.

MattBrittan:
I think a lot of this comes down to a discussion around the desired level of functionality in the `paho` package. Currently `paho` accepts a pre-populated `net.Conn` and runs from there (i.e. sends `CONNECT` packet, fully terminates when connection is lost). I really like this design because it keeps all of the connection management stuff separate (this caused a lot of issues in the 3.1.1 client due to confusing interactions between the various running processes).

If `paho` accepts a pre-connected `net.Conn` (meaning that it cannot reconnect) then I feel that managing persistence of the session state should also be out of scope.  Of course `paho` needs to interact with the state (flagging that a `PUBLISH` has been sent, the receipt of `ACK` packets etc) and my current thought is that this could be handled by an extended version of `MIDService` (do need to do some experimentation to check that this will work).

Doing things this way keeps `paho` really simple. If someone wants a really basic MQTT client (i.e. for embedded use) then they can just use `paho` (perhaps using `autopaho` as documentation on best practice use). Whereas if a user wants a full client then `autopaho` will offer that (with a small selection of common options), or they can fork `autopaho` if they want to do something unusual (a big issue with the 3.1.1 client was that it tried to cater for every requirement).

So, with the above in mind, I feel that the task of loading the session state (to pre-populate `MIDService`) should be managed outside of `paho`; hence my comments around there being no need for `paho` to mark of Packet IDs as in-use (because this would also be done externally). This really all comes back to the big question, "how do we handle session persistence", and I'm reluctant to implement functionality that *may* be needed to support this (because once users have used a feature its difficult to remove/change it later - don't want to break end users code!).

I'll try to put some time over the next couple of weeks into fleshing out how the store might work; perhaps it might be worth having a discussion around this (hopefully you have had a chance to talk to Al?).

vishnureddy17:
> If paho accepts a pre-connected net.Conn (meaning that it cannot reconnect) then I feel that managing persistence of the session state should also be out of scope. Of course paho needs to interact with the state (flagging that a PUBLISH has been sent, the receipt of ACK packets etc)

Good point.

> my current thought is that this could be handled by an extended version of MIDService (do need to do some experimentation to check that this will work).

Why MIDService and not the persistence implementation (however that'd look like)? I'm wondering if you have a good mental model for what should belong in the packet ID tracker vs. the persistent session store. It might clarify my thinking.

> I'll try to put some time over the next couple of weeks into fleshing out how the store might work; perhaps it might be worth having a discussion around this (hopefully you have had a chance to talk to Al?).

Happy to have a discussion on this on or off GitHub—you have my e-mail address. Yep, I've talked to Al!

MattBrittan:
>Why MIDService and not the persistence implementation (however that'd look like)?

Just a thought at this point. The lifecycle of a MID matches that of an outgoing message (allocated when message created, released after appropriate acknowledgments received) so it may make sense to combine the two.

MattBrittan:
I'll close this one off as I believe PR #172 addresses the requirement in a different way (and significantly changes how message IDs are handled meaning this PR is probably no longer relevant).

PR #142, opened 2023-06-22
Handle Orphaned Operations

vishnureddy17:
related issue: #134

Here's a proposal for a design to handle orphaned operations (PUBLISH w/ QoS >0, SUBSCRIBE, and UNSUBSCRIBE). 

This would allow the user to get back the result of an operation after a timeout. More importantly, this allows us to kick off a PUBLISH operation in the background when connecting with an existing session comes into the picture.

It seems to work as intended, and I'm happy to write some tests for it. However, I wanted to get feedback on this before I go through the effort since this is a substantial change and I'm not sure if this is the direction the project would be willing to take.

The basic idea is that PUBLISH w/ QoS >0, SUBSCRIBE, and UNSUBSCRIBE are handled inside an internal goroutine that puts the result in a channel when it is received from the server. The user-facing Publish(), Subscribe(), and Unsubscribe() functions call the internal functions. If the context is cancelled before the internal goroutines return a result on the channel, the user-facing function returns with an error and the operation is still tracked in the background. There are user-configurable callbacks to get the results for orphaned operations.

### DIFF
--- a/paho/client.go
+++ b/paho/client.go
@@ -710,7 +710,10 @@ func (c *Client) Subscribe(ctx context.Context, s *Subscribe) (*Suback, error) {
 		c.debug.Printf("adding SUBSCRIBE for %d to OrphanedOps due to context cancellation: %w", sp.PacketID, subCtx.Err())
 		c.OrphanedOps.AddOp(resp)
 		return nil, subCtx.Err()
-	case r := <-resp:
+	case r, ok := <-resp:
+		if !ok {
+			return nil, fmt.Errorf("SUBSCRIBE response channel closed")
+		}
 		return SubackFromPacketSuback(r.Content.(*packets.Suback)), nil
 	}
 }
@@ -776,7 +779,10 @@ func (c *Client) Unsubscribe(ctx context.Context, u *Unsubscribe) (*Unsuback, er
 		c.debug.Printf("adding UNSUBSCRIBE for %d to OrphanedOps due to context cancellation: %w", up.PacketID, unsubCtx.Err())
 		c.OrphanedOps.AddOp(resp)
 		return nil, unsubCtx.Err()
-	case r := <-resp:
+	case r, ok := <-resp:
+		if !ok {
+			return nil, fmt.Errorf("UNSUBSCRIBE response channel closed")
+		}
 		return UnsubackFromPacketUnsuback(r.Content.(*packets.Unsuback)), nil
 	}
 }
@@ -830,7 +836,10 @@ func (c *Client) Publish(ctx context.Context, p *Publish) (*PublishResponse, err
 	case <-pubCtx.Done():
 		c.OrphanedOps.AddOp(resp)
 		return nil, pubCtx.Err()
-	case r := <-resp:
+	case r, ok := <-resp:
+		if !ok {
+			return nil, fmt.Errorf("PUBLISH response channel closed")
+		}
 		switch r.Type {
 		case packets.PUBACK:
 			return PublishResponseFromPuback(r.Content.(*packets.Puback)), nil
--- a/paho/client.go
+++ b/paho/client.go
@@ -40,6 +40,7 @@ type (
 		// wrapper or extend the custom net.Conn struct with sync.Locker.
 		Conn          net.Conn
 		MIDs          MIDService
+		OrphanedOps   OrphanedOps
 		AuthHandler   Auther
 		PingHandler   Pinger
 		Router        Router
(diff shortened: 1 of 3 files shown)


paho/client.go
```diff
@@ -641,133 +655,140 @@ func (c *Client) Subscribe(ctx context.Context, s *Subscribe) (*Suback, error) {
...
-	defer cf()
-	cpCtx := &CPContext{subCtx, make(chan packets.ControlPacket, 1)}
+	respInternal := make(chan packets.ControlPacket)
+	resp := make(chan packets.ControlPacket, 1)
 
 	sp := s.Packet()
 
-	mid, err := c.MIDs.Request(cpCtx)
+	var err error
+	sp.PacketID, err = c.MIDs.Request(&CPContext{context.TODO(), respInternal})
```
vishnureddy17, line 664: I think that contexts should be removed from the packet ID tracker, but I put `context.TODO()` as a placeholder for now.

paho/client.go
```diff
@@ -791,79 +812,154 @@ func (c *Client) Publish(ctx context.Context, p *Publish) (*PublishResponse, err
...
+	resp := make(chan packets.ControlPacket, 1)
+
+	respInternal := make(chan packets.ControlPacket)
+
+	var err error
+	if pb.PacketID == 0 {
+		pb.PacketID, err = c.MIDs.Request(&CPContext{context.TODO(), respInternal})
+	} else {
+		// ClaimID() is not yet implemented
+		err = fmt.Errorf("publishing with a specific PacketID is not yet implemented")
```
vishnureddy17, line 858: This is here as a placeholder, but this is intended to handle the case where we are re-delivering a PUBLISH when resuming an existing session

paho/client.go
```diff
@@ -641,133 +655,140 @@ func (c *Client) Subscribe(ctx context.Context, s *Subscribe) (*Suback, error) {
...
-	switch {
-	case len(sa.Reasons) == 1:
-		if sa.Reasons[0] >= 0x80 {
-			var reason string
-			c.debug.Println("received an error code in Suback:", sa.Reasons[0])
-			if sa.Properties != nil {
-				reason = sa.Properties.ReasonString
-			}
-			return sa, fmt.Errorf("failed to subscribe to topic: %s", reason)
-		}
```
vishnureddy17, line 690: In my implementation, I do not return an error if the server returns an error reason code. I made the same change for Unsubscribe() and Publish()

IMO, it should not be considered an error from the client perspective, it is just one of the possible results from a successfully completed operation. Curious to hear thoughts on this.

paho/client.go
```diff
@@ -641,133 +655,140 @@ func (c *Client) Subscribe(ctx context.Context, s *Subscribe) (*Suback, error) {
...
-	}
+		resp <- receivedPacket
+	}()
 
-	return sa, nil
+	select {
+	case <-subCtx.Done():
+		c.debug.Printf("adding SUBSCRIBE for %d to OrphanedOps due to context cancellation: %w", sp.PacketID, subCtx.Err())
+		c.OrphanedOps.AddOp(resp)
+		return nil, subCtx.Err()
```
vishnureddy17, line 712: Important detail: if the context is cancelled, the channel gets added to `OrphanedOps`, which holds onto the channels for orphaned operations and calls the user-specified callback if a result comes in. Same deal with Unsubscribe() and Publish()

paho/client.go
```diff
@@ -791,79 +812,154 @@ func (c *Client) Publish(ctx context.Context, p *Publish) (*PublishResponse, err
...
+				reasonString := fmt.Sprintf(
+					"expected PUBACK for %d, got %s",
+					pb.PacketID,
+					receivedPacket.PacketType(),
+				)
+				c.errors.Printf("disconnecting due to server protocol error. %s", reasonString)
+				go c.Disconnect(&Disconnect{
+					ReasonCode: packets.DisconnectProtocolError,
+					Properties: &DisconnectProperties{ReasonString: reasonString},
+				})
```
vishnureddy17, line 902: If we get a protocol error, the client disconnects with the Protocol Error reason code. Based on my understanding of the MQTT spec, this is the correct behavior, but it wasn't being done before.

vishnureddy17:
After working with this design for a bit, I actually think this is not a good approach. By putting each publish into its own goroutine it is difficult to guarantee ordering that the MQTT spec requires.

vishnureddy17:
Going to close this, I don't think this is a good path to go down.

Issue #143, opened 2023-06-30
Proxy support


Lorderot:
**Is your feature request related to a problem? Please describe.**
I want to be able to connect to the MQTT Broker via proxy.

**Describe the solution you'd like**
One of the following:
- The same as in paho.mqtt.golang repository: https://github.com/eclipse/paho.mqtt.golang/blob/master/netconn.go#L85
- Making it possible to configure dialer

Lorderot:
We've switched from paho.mqtt.golang to paho.golang (mqtt3 -> mqtt5) and we've faced the issue that suddenly new library doesn't regard "all_proxy" setting, while the previous library did.

PR #144, opened 2023-06-30
[TRIVIAL]: Add proxy support

Lorderot:
Trivial change to respect "all_proxy" env var, as it was done in paho.mqtt.golang library.
Taken from:
https://github.com/eclipse/paho.mqtt.golang/blob/master/netconn.go#L85


Lorderot:
just an example of the change for https://github.com/eclipse/paho.golang/issues/143

alsm:
Thanks for this, useful addition. Needs to pass the ECA before I can merge it.

Lorderot:
@alsm thanks, I've signed the ECA.

Lorderot:
hi @alsm, can we proceed with this? thanks!

Issue #145, opened 2023-07-26
0 Packet Identifier in acknowledgments causes panic


MattBrittan:
**Describe the bug**
When acknowledgments are received `c.MIDs.Get(recv.PacketID())` is run. If the PacketID is 0 this leads to a panic (due to `return m.index[i-1]` in `message_ids.go`.

**To reproduce**
Send a PUBACK with 0 packet ID (occurred whilst I was testing something else). Quick and dirty test to replicate:

```
func TestClientReceiveQoS10ClientID(t *testing.T) {
	rChan := make(chan struct{})
	ts := newTestServer()
	go ts.Run()
	defer ts.Stop()

	c := NewClient(ClientConfig{
		Conn: ts.ClientConn(),
		Router: NewSingleHandlerRouter(func(p *Publish) {
			assert.Equal(t, "test/0", p.Topic)
			assert.Equal(t, "test payload", string(p.Payload))
			assert.Equal(t, byte(0), p.QoS)
			close(rChan)
		}),
	})
	require.NotNil(t, c)
	c.SetDebugLogger(log.New(os.Stderr, "RECEIVEQOS0: ", log.LstdFlags))
	c.SetDebugLogger(log.New(os.Stderr, "RECEIVEQOS1: ", log.LstdFlags))

	c.serverInflight = semaphore.NewWeighted(10000)
	c.clientInflight = semaphore.NewWeighted(10000)
	c.stop = make(chan struct{})
	c.publishPackets = make(chan *packets.Publish)
	go c.incoming()
	go c.PingHandler.Start(c.Conn, 30*time.Second)
	go c.routePublishPackets()

	err := ts.SendPacket(&packets.Puback{
		PacketID:   0,
		ReasonCode: 0,
	})
	require.NoError(t, err)
	time.Sleep(time.Second)
}
```

**Debug output**

```
=== RUN   TestClientReceiveQoS10ClientID
RECEIVEQOS1: 2023/07/26 23:26:34 received PUBACK packet with id 0
RECEIVEQOS1: 2023/07/26 23:26:34 client stopping, incoming stopping
panic: runtime error: index out of range [65535] with length 65535

goroutine 8 [running]:
github.com/eclipse/paho.golang/paho.(*MIDs).Get(0xc00005e4b0?, 0xb333?)
	C:/Users/Matt/Code/Temp/pahotemp/paho.golang/paho/message_ids.go:87 +0xcc
[...]

PR #147, opened 2023-07-27
handling 0 Packet Identifier in MIDs

tomatod:
This is related to #145 . 

To avoid panic, handling 0 Packet Identifier although it is invalid (valid range is 1 - 65535).


MattBrittan:
Any chance that you could add a test? (I provided a very rough one in the issue). I believe that MIDs is going to change further (see existing issues etc) so having a test for this situation would be beneficial in preventing the reintroduction of this problem.

tomatod:
@MattBrittan
Thank you for comments. I added some tests for MIDs.Get() and Free()

PR #153, opened 2023-08-03
Add CONNACK packet to autopaho OnConnectError func

### LINKED ISSUE #152: Access to CONNACK packet in autopaho
If establishing a connection with autopaho fails, we get a descriptive error from Properties.ReasonString. It would be more useful if we also got the ReasonCode, as that would enable us to reliably handle the error.

For example, if the server returns reason code 0x87, we now know we have to set up fresh access credentials. 

Change the [connect error handling function](https://github.com/eclipse/paho.golang/blob/master/autopaho/auto.go#L48) to `OnConnectError(*paho.Connack, error)`

We currently handle this by parsing the error, but this is unreliable, especially when connecting to multiple MQTT server implementations.

**Additional context**
This would be a breaking change in autopaho ClientConfig.

thejan2009:
This enables handling connection errors based on response code received by the server.

Fixes #152

### DIFF
--- a/autopaho/auto.go
+++ b/autopaho/auto.go
@@ -52,7 +52,7 @@ type ClientConfig struct {
 	AttemptConnection func(context.Context, ClientConfig, *url.URL) (net.Conn, error)
 
 	OnConnectionUp func(*ConnectionManager, *paho.Connack) // Called (within a goroutine) when a connection is made (including reconnection). Connection Manager passed to simplify subscriptions.
-	OnConnectError func(error)                             // Called (within a goroutine) whenever a connection attempt fails
+	OnConnectError func(error)                             // Called (within a goroutine) whenever a connection attempt fails. Will wrap autopaho.ConnackError on server deny.
 
 	Debug      paho.Logger // By default set to NOOPLogger{},set to a logger for debugging info
 	PahoDebug  paho.Logger // debugger passed to the paho package (will default to NOOPLogger{})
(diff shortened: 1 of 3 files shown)


autopaho/examples/docker/publisher/main.go
```diff
@@ -27,8 +27,10 @@ func main() {
 		KeepAlive:         cfg.keepAlive,
 		ConnectRetryDelay: cfg.connectRetryDelay,
 		OnConnectionUp:    func(*autopaho.ConnectionManager, *paho.Connack) { fmt.Println("mqtt connection up") },
-		OnConnectError:    func(err error) { fmt.Printf("error whilst attempting connection: %s\n", err) },
-		Debug:             paho.NOOPLogger{},
+		OnConnectError: func(ca *paho.Connack, err error) {
+			fmt.Printf("error code %d whilst attempting connection: %s\n", ca.ReasonCode, err)
```
MattBrittan, line 31: This will panic if `ca` is `nil` (e.g. when the network connection failed)

MattBrittan (commented):
This would be a breaking change (which is OK as we are pre v1 but still something we try to avoid unless it's really necessary). 

I wonder whether an alternative option would be to return a custom error; the user could then use [`errors.As`](https://pkg.go.dev/errors#As`) to extract the information. Personally I think this may be tidier given that most of the time it's likely that the `*paho.Connack` will be `nil` (I'd expect a lot more network issues than rejected CONNECT packets).

Thoughts?

MattBrittan:
*Confirmed ECA on file*.

Thanks very much for the PR (and picking up the additional potential panic due to `nil` callback) - looks good.

PR #158, opened 2023-08-08
Autopaho: Call error call back a maximum of once per connection

### LINKED ISSUE #157: Useless "use of closed network connection" error after autopaho disconnect
**Describe the bug**
While working on #153, I started finding new errors in the log.

On latest master, I reliably get autopaho `OnClientError` even after legitimately disconnecting the client, the error says something like `read tcp 10.42.1.27:48018->18.196.199.37:443: use of closed network connection`

**To reproduce**
Create autopaho client with defined OnClientError, then connect and disconnect.

**Expected behaviour**
No error.

**Software used:**
I suspect this is the culprit https://github.com/eclipse/paho.golang/commit/c5d83b82693d45a7f9c8265fbbb3a4722ffa1f02/autopaho/auto.go#diff-c1ef7c1c120786f08491971c56e70b874f779e875fc0352a4085b3a9954a754cR195

What probably happened before is that the goroutine that sent this error was canceled before it could send the error through unbuffered channel. Now the channel is buffered so the error always arrives.

CC @MattBrittan

MattBrittan:
autoPaho had the potential to call the error call beck multiple times in relation to one event.
This mirrored the `paho` functionality but does not make much sense from an end user perspective,
so this change prevents multiple calls to `OnClientError` in relation to a single connection.
We also prevent errors from being logger when the user has called `Disconnect` (it's not an error
if it's the result of a request) and add a test.

closes #157

### DIFF
--- a/autopaho/error.go
+++ b/autopaho/error.go
@@ -58,10 +58,9 @@ func (e *errorHandler) handleError(err error) bool {
 		e.debug.Printf("received error: %s", err)
 		errChan <- err
 		return true
-	} else {
-		e.debug.Printf("received extra error: %s", err)
-		return false
 	}
+	e.debug.Printf("received extra error: %s", err)
+	return false
 }
 
 // DisconnectError will be passed when the server requests disconnection (allows this error type to be detected)
--- a/autopaho/error.go
+++ b/autopaho/error.go
@@ -10,6 +10,9 @@ import (
 // errorHandler provides the onClientError callback function that will be called by the Paho library. The sole aim
 // of this is to pass a single error onto the error channel (the library may send multiple errors; only the first
 // will be processed).
+// The callback userOnClientError will be called a maximum of one time. If userOnServerDisconnect is called, then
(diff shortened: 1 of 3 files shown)


autopaho/error.go
```diff
@@ -39,15 +48,19 @@ func (e *errorHandler) onServerDisconnect(d *paho.Disconnect) {
...
+	e.mu.Unlock()
+	if errChan != nil {
 		e.debug.Printf("received error: %s", err)
-		e.errChan <- err
-		e.errChan = nil
+		errChan <- err
+		return true
 	} else {
 		e.debug.Printf("received extra error: %s", err)
+		return false
```
thejan2009, line 63: Nitpick, we can now avoid `else` clause because of return statements.

thejan2009 (approved):
This seems reasonable, given that `paho` will close the connection on error. I've tested the commit in two different applications and observed no more bogus "use of closed network connection" errors.

PR #172, opened 2023-10-19
Implement session state (full QOS1/2 support) and publish queue

### LINKED ISSUE #25: Persistence is not implemented

MattBrittan:
This change adds support for Session State which enables full QOS1/2 support (with information about messages that are not completely acknowledged being stored in memory or on disk). Adding this functionality required major changes to the library, but most users' code should run without alteration (see the updated readme for further info).

I have tested this code, but there will be bugs (and a few todos remain). I feel that it is good enough to be pulled into @master so others can test (and provide fixes).

closes  #25


alsm:
Thanks for working on this @MattBrittan a big improvement that takes us that much closer to a v1 release

Issue #168, opened 2023-09-29
Consider changing the approach used for publish handlers


vishnureddy17:
_Continuing the discussion from https://github.com/vishnureddy17/paho.golang/pull/19#discussion_r1337408392_

I feel it would be better to get rid of the user-defined routers in `paho` and instead just provide the following methods directly on the paho client:
`func (c *Client) SetHandlerAllPublishes(func(*Publish))`
`func (c *Client) UnsetHandlerAllPublishes()`
`func (c *Client) SetTopicHandler(topic string, func(*Publish))`
`func (c *Client) UnsetTopicHandler(topic string)`

This would still allow the same use cases as before.

Currently, the behavior for registering handlers differs depending on the router implementation being used, and this change would make things more consistent.

cc @ashtonian @MattBrittan

MattBrittan:
I don't really have a big issue with the existing setup. It's relatively simple to use and provides a lot of flexibility (I would be tempted to move the router into a sub-package to fully separate it).

>Currently, the behavior for registering handlers differs depending on the router implementation

The process for creating a handler instance does vary (i.e. `NewStandardRouter()` vs `NewSingleHandlerRouter(h MessageHandler)`) but is this really an issue? (if you want a single handler router, then setting the handler at the time of creation makes sense as there will be one, and only one, handler).

One issue with the proposed interface is reconnections. Do we provide a way to retrieve the existing router config and pass it into a new instance of the client? (ideally `autopaho` should be able to reconnect without any need to touch the router config).

I also feel that adding this into `paho` locks in some functionality that may make sense in many cases, but will sometimes surprise users e.g.:
- Does adding a handler for a topic mean that the new handler will be called as well as, or instead of, existing handlers.
- `UnsetTopicHandler` may unset multiple handlers (which could be confusing if data from a topic is used in separate packages).

In some ways I think going the other way might be preferable. By this I mean simplifying `paho` such that `Router` is replaced with a callback (i.e. ` func(*packets.Publish)` or perhaps `func(*packets.Publish) error)`. My rationale for this is that the `paho` library only calls a single function in the `Router` (`Route(*packets.Publish)`) so it's really not necessary for us to dictate the router structure (and the current structure requires `SingleHandlerRouter` to implement `RegisterHandler and just ignore the topic!). Having said that, the advantage of the current setup is that if you have a `Client` you know you can call `client.Router.RegisterHandler` (and this simplifies things for new users).

vishnureddy17:
[...]

PR #184, opened 2023-11-04
Add defaultHandler to standard router

MattBrittan:
Adds a default Handler to StandardRouter which would simplify removing SingleHandlerRouter.
This is also useful to have (I generally use it to log any unexpected messages as it means
there is probably a bug!).

ref #168


vishnureddy17:
This is great, thanks!

PR #186, opened 2023-11-08
Remove SingleHandlerRouter

MattBrittan:
SingleHandlerRouter did not adhere to the
requirements set out in the router interface
documentation.

Have retained NewSingleHandlerRouter so that
most code should run without modification.

ref #168


MattBrittan:
Accepting this as it's been discussed and I'm keen to update my live code to utilise it.

Issue #160, opened 2023-08-12
Manual ACK's, session state, and connection reestablishment


MattBrittan:
Whilst working on session state persistence I believe (happy to be proven wrong!) I have come across a small issue with the design of the `client` API so am raising it for discussion. I'm not sure how frequently manual acknowledgements are used (@fracasula this may be one for you?). Note that this is not currently an issue with `autopaho` because it does not support manual acknowledgements (there is no way to call `ACK`). 

While this is a fairly niche issue I think it's important that we come up with a solution before v1 because I believe an API change will be required.

Consider the following situation:

 * Manual ACK enabled
 * QOS1+ message received and passed to client
 * Connection drops and is re-established
 * `client.ACK()` called (either on the old or the new `client`)

There are issues here regardless of whether the session state survives the reconnection:
  * If the session is deleted then we may be acknowledging the wrong message (the broker may have reused the ID in the interim)
  * If the session state remains then it's not really clear what action to take (the message may well have been retransmitted and the library has no way of knowing if the client will eventually acknowledge it potentially leading to a duplicate ACK).

As mentioned above, the user may call `ACK` on the original `Client` or the new one; there are issues in both cases:

 * Client that dropped - The ACK cannot be sent (due to connection loss) and we cannot reliably update the session state (because the server may have retransmitted the message). 
 * New Client - If the session was not retained (e.g. due to `CleanSession`) then sending the ACK may acknowledge a different message (as the server is free to reuse the ID).

I guess we could just say "this is the users problem - they should use a waitgroup to track handlers and only reconnect when they have all completed" but I'm not sure that is the best option....

I cannot see a good way of handling this currently because there is currently no reliable way of working out which connection the `ACK` relates to.  A couple of ways this could be solved are:

* Add a `NoAck` (or similar) function that indicates that the message will not be acknowledged. Then provide a way for the library to tell when all handlers have completed (i.e. either `Ack` or `NoAck` have been called for all QOS1+ publish messages passed to the user) and only reconnect after that time. 
* Use the V3 client approach where the `struct` passed to `Router.Route()` implements `Ack()` and the user calls that instead of `client.Ack()`. This would enable the library to prevent the `Ack` from being sent if the session was cleared (not a perfect solution as it's not really safe to assume a QOS1 resend is a duplicate).

[...]

PR #187, opened 2023-11-08
keepalive=0 -> panic, reject empty brokers

MattBrittan:
paho would panic if keepalive was 0.
Autopaho would go into infinite loop if
no servers supplied.

ref #160


MattBrittan:
Closing this one as the PR's got a bit mixed up!

PR #189, opened 2023-11-09
Accept 0 keepalive

MattBrittan:
paho would panic if keepalive was 0.


MattBrittan:
Cannot a reason why this should not be merged.

Issue #192, opened 2023-11-11
Trying to log into server with wrong credentials times out instead of returning an error


XANi:
when  trying to connect to server
```go
	conn, err := autopaho.NewConnection(t.mqttCtx, t.mqttCfg)
	if err != nil {
		return err
	}
	connectCtx, _ := context.WithTimeout(t.mqttCtx, time.Second*30)
	if err = conn.AwaitConnection(connectCtx); err != nil {
		return fmt.Errorf("error connecting to mq: %w", err)
	}
```

and password is wrong (or unset, btw why it doesn't take password from url.URL automatically but need separate call?) this will time out instead of returning authorization error. I've snooped on tcpdump and  server is returning code 135 not authorized to the client but client just hangs till timeout.

Version: `v0.12.0`

MattBrittan:
This is by design; `autopaho` will just keep attempting to connect until you stop it (by cancelling the context or calling `Disconnect`). `AwaitConnection` is just a utility function to allow you to easily detect when/if the connection comes up. `AwaitConnection` was mainly added to simplify examples; I would not expect  to be used all that often in production code (it's generally better to use the [`OnConnectionUp`](https://pkg.go.dev/github.com/eclipse/paho.golang@v0.12.0/autopaho#ClientConfig) callback because the connection may drop/be re-established at any time).

The aim of `autopaho` establish/maintain a connection in whatever way it can, so it will retry until you stop it (if you want to manage the connection yourself then use `paho`). In many production environments there is little you can do when there is a connection issue (perhaps report it to some logging server) and retrying constantly is the right approach. We don't want to make assumptions about what action should be taken if there is an error (even in your case this issue could be caused by an issue on the server so retrying may eventually lead to a connection).

If you want to handle such errors then add an [`OnConnectError`](https://pkg.go.dev/github.com/eclipse/paho.golang@v0.12.0/autopaho#ClientConfig) handler and take whatever action is needed to handle the issue.

>btw why it doesn't take password from url.URL automatically but need separate call?

Partially because it can be confusing; for example when establishing a web-socket connection should the username/password be passed when establishing the http connection (e.g. `Authorization` header), in the MQTT connect packet, or both?.

XANi:
>This is by design; autopaho will just keep attempting to connect until you stop it (by cancelling the context or calling Disconnect). AwaitConnection is just a utility function to allow you to easily detect when/if the connection comes up. 

[...]

PR #206, opened 2023-11-19
pass `*url.URL` to ConnectPacketBuilder

XANi:
As per discussion in #192 I've added URL as a parameter, and also an example on how to use it to extract password from URL

### DIFF
--- a/autopaho/auto.go
+++ b/autopaho/auto.go
@@ -79,7 +79,7 @@ type ClientConfig struct {
 	WillMessage    *paho.WillMessage
 	WillProperties *paho.WillProperties
 
-	ConnectPacketBuilder func(*paho.Connect) *paho.Connect // called prior to connection allowing customisation of the CONNECT packet
+	ConnectPacketBuilder func(*paho.Connect, *url.URL) *paho.Connect // called prior to connection allowing customisation of the CONNECT packet
 
 	// DisconnectPacketBuilder - called prior to disconnection allowing customisation of the DISCONNECT
 	// packet. If the function returns nil, then no DISCONNECT packet will be passed; if nil a default packet is sent.
@@ -158,7 +158,9 @@ func (cfg *ClientConfig) SetWillMessage(topic string, payload []byte, qos byte, 
 // Deprecated: Set ConnectPacketBuilder directly instead. This function exists for
 // backwards compatibility only (and may be removed in the future).
 func (cfg *ClientConfig) SetConnectPacketConfigurator(fn func(*paho.Connect) *paho.Connect) bool {
-	cfg.ConnectPacketBuilder = fn
+	cfg.ConnectPacketBuilder = func(pc *paho.Connect, u *url.URL) *paho.Connect {
+		return fn(pc)
+	}
 	return fn != nil
 }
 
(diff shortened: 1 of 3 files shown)


autopaho/auto.go
```diff
@@ -79,7 +79,7 @@ type ClientConfig struct {
 	WillMessage    *paho.WillMessage
 	WillProperties *paho.WillProperties
 
-	ConnectPacketBuilder func(*paho.Connect) *paho.Connect // called prior to connection allowing customisation of the CONNECT packet
+	ConnectPacketBuilder func(*paho.Connect, *url.URL) *paho.Connect // called prior to connection allowing customisation of the CONNECT packet
```
MattBrittan, line 82: I wonder if we should pass a `struct` instead? Trying to think of any other data that might be of use to `ConnectPacketBuilder` in the future (using a `struct` would mean future additions would be non-breaking) but cannot currently think of anything else that would be useful...

autopaho/auto.go
```diff
@@ -157,7 +157,7 @@ func (cfg *ClientConfig) SetWillMessage(topic string, payload []byte, qos byte,
 //
 // Deprecated: Set ConnectPacketBuilder directly instead. This function exists for
 // backwards compatibility only (and may be removed in the future).
-func (cfg *ClientConfig) SetConnectPacketConfigurator(fn func(*paho.Connect) *paho.Connect) bool {
+func (cfg *ClientConfig) SetConnectPacketConfigurator(fn func(*paho.Connect, *url.URL) *paho.Connect) bool {
```
MattBrittan, line 160: I don't think we should change this; it's an unnecessary breaking change to a depreciated function. Just wrap the function the user passed.

autopaho/auto.go
MattBrittan: Probably fine as is (unless you can think of other data that could be useful).

XANi: It could just be passed entire `ClientConfig`, although only other useful thing to configure before connection that I can think of is TLS configuration and it's too late for that in buildConnectPacket

autopaho/auto.go
XANi: fixed

autopaho/auto.go
XANi: But URL is really only thing different between each connection attempt so any other external info user might want in that function could just be provided by the user when defining the function. Nothing really stops hook from looking like this
```go
config := autopaho.ClientConfig{
		ServerUrls:        []*url.URL{serverURL},
		ConnectRetryDelay: 5 * time.Second,
		ConnectTimeout:    5 * time.Second,
			ClientID: "test",
		},
	}
config.ConnectPacketBuilder = func(c *paho.Connect, u *url.URL) *paho.Connect {  
    doSomethingWith(config)
}

MattBrittan: Yea, something to do with the TLS connection (or the `net.Conn` itself) was what I was thinking of. However having thought of this I could not come up with a real use-case...

XANi: Well the simplest one would be different server cert/client cert/ca per server but if ever someone had that use case `AttemptConnection` hook can cater for that just fine. 

I guess one thing that could be improved is making `attemptXXXConnection()` functions public so someone making custom `AttemptConnection` hook could reuse it, I can add a commit changing it here if you want.

MattBrittan: Thanks - however I feel that the `attemptXXXConnection` functions are pretty simple & generic (also easy to copy/modify) so it's best to leave them private (making them public may make future changes more difficult because changing a public interface = breaking change).

XANi: Yeah I suggested it because the `packets.NewThreadSafeConn()` in `attemptTLSConnection` might be not very obvious on first look but the comment on callback itself explains it well enough

PR #211, opened 2023-11-22
Replace Router with OnPublishReceived Callback

MattBrittan:
This demonstrated how Router could be replaced in a mostly non-breaking way.
Note that I have not worked through all examples to test (know that autopaho/rpc will be broken).

reference #168

### DIFF
--- a/paho/client.go
+++ b/paho/client.go
@@ -32,7 +32,15 @@ var (
 )
 
 type (
-	// ClientConfig are the user configurable options for the client, an
+	PublishReceived struct {
+		Packet *Publish
+		Client *Client // The Client that received the message (note that the connection may have been lost post-receipt)
+
+		AlreadyHandled bool    // Set to true if a previous callback has returned true (indicating some action has already been taken re the message)
+		Errs           []error // Errors returned by previous handlers (if any).
+	}
+
+	// ClientConfig are the user-configurable options for the client, an
 	// instance of this struct is passed into NewClient(), not all options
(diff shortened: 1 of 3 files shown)


paho/client.go
```diff
@@ -142,9 +169,20 @@ func NewClient(conf ClientConfig) *Client {
 	if c.PacketTimeout == 0 {
 		c.PacketTimeout = 10 * time.Second
 	}
-	if c.Router == nil {
-		c.Router = NewStandardRouter()
+
+	if c.Router == nil && len(c.onPublishReceived) == 0 {
+		c.Router = NewStandardRouter() // Maintain backwards compatibility (for now!)
```
vishnureddy17, line 174: I'm wondering if it would be worth just avoiding the backward compatibility.

I understand that it's nice to avoid breaking users, but I'd also be wary of adding extra code paths for the sake of backward compatibility before this even comes out of beta. Beta is the chance to try things out and make changes when things don't work.

vishnureddy17 (approved):

paho/client.go
MattBrittan: @vishnureddy17 not really sure either way on this one. I think I'll push this change in and we can then look at removing the Router in a subsequent change (doing it this way means I don't need to update all of the examples etc in order to get this change in). We could remove the `Router` before the next release or, alternatively, wait for the following release (allowing users to update to the latest release and then remove references to the depreciated Router over a month or so).

Issue #214, opened 2023-12-25
Incorrect assumption about PUBLISH errors in autopaho can cause indefinite head-of-queue looping


vishnureddy17:
In `autopaho`, `managePublishQueue` assumes that if `paho.PublishWithOptions()` returns an error that is _not_ `paho.ErrNetworkErrorAfterStored`, the error is either temporary or the connection will drop: 
https://github.com/eclipse/paho.golang/blob/a6def521ee1aecf0b1cef6bfc1a7022102457807/autopaho/auto.go#L531-L553

However, this is not always the case. All the errors returned in the code snippet below for `paho.PublishWithOptions()` are neither temporary nor imply a pending disconnection: 
https://github.com/eclipse/paho.golang/blob/a6def521ee1aecf0b1cef6bfc1a7022102457807/paho/client.go#L748-L762

As a result, `managePublishQueue` will loop indefinitely on the first entry in the queue in these cases. 

**Proposed Solution**
To solve this, I propose creating a custom `error` type called `PahoArgumentError`, which will be returned by `paho.PublishWithOptions()` in these cases (this `error` type will likely be relevant elsewhere in `paho`). Then `autopaho` can do a type assertion and quarantine any publish in the queue that returned `PahoArgumentError`.

MattBrittan:
Agreed; I did consider doing something similar when introducing this, but felt that it was OK for an initial PR (hard to know when to stop and submit the PR). My thought was something like `paho.FatalError` (indicating that it's pointless to retry) because this could also apply if the broker returns something like "Payload format invalid"). Happy with either option (will try to get a bit more work done on this over the next couple of days).

vishnureddy17:
That sounds good, I also think it's worth returning such an error from Subscribe and Unsubscribe, which would be used to distinguish between errors that are due to a client failure/disconnection and errors that are due to something fatally wrong with the subscribe/unsubscribe

MattBrittan:
I've had an initial go at this; went with your solution (it's probably useful to be able to differentiate between
a range of "fatal errors"). Ideally this would be extended to look at broker response codes (but I don't think that 
is as urgent).

vishnureddy17:
closing due to #226, @MattBrittan feel free to reopen if you still think it's important to keep around

vishnureddy17:
> Ideally this would be extended to look at broker response codes (but I don't think that
is as urgent).

This is really only doable if #216 is addressed. 

[...]

Issue #216, opened 2023-12-29
Receive notification of acknowledgments when using async publish


MattBrittan:
**Is your feature request related to a problem? Please describe.**

When using async PUBLISH (i.e. `PublishWithOptions` with `PublishMethod_AsyncSend`) it's not currently possible to determine when the message has been acknowledged (with QOS 2 messages it would be useful to know when both steps are complete).

**Describe the solution you'd like**
This was discussed a bit in issue #207 options include:

> One way to achieve this is to change AddToSession() in SessionManager to return a function that will block until the ack is received. The paho client would just return that function to the user. The function would be valid for use for as long as the session is.

>My thought on this was to add callbacks to PublishOptions; we could then store the PublishOptions in state.clientPackets (making it easy to call when an associated event occurs). I prefer the callback approach because I think it's easier for the end user (and avoids the need to have a lot of go-routines hanging around waiting for acknowledgements). One note of caution is that the session may outlast the client instance (and I can't think of a good way of handling that!).

MattBrittan:
Carrying on discussion from #207 comment by @vishnureddy17 

>> One note of caution is that the session may outlast the client instance (and I can't think of a good way of handling that!).

>I think that's expected and necessary. As a user, I'd want to know when the publish was handled in the session (regardless of how many reconnections happened in the interim). In the solution you suggest, that would mean that the callback can potentially be called long after a particular client is dead.

Just to clarify; by "client instance" I meant the users program (rather than an instance of `paho`/`autopaho`). As the session can be stored on disk it's quite possible for the following to happen:

1. User publishes message and it's added into session (or queue in autopaho)
2. Client application restarts (say a server restart or application update)
3. Message actually sent

This raises a few questions:
* Is the callback called when the application is shutting down (only possible on a clean shutdown obviously)?
* Does the library provide a way to 'reregister' the callbacks (or whatever mechanism is used) when starting up?

My thought was always that if autopaho is in use then whatever notification method is selected should survive reconnections (these should be mostly hidden from the end user). We do also need to handle the situation where the session is lost (either due to `Clean Start` or the session expiring) so probably want to be able to pass an error to the user when this happens.

vishnureddy17:
> Does the library provide a way to 'reregister' the callbacks (or whatever mechanism is used) when starting up?

[...]

PR #226, opened 2024-01-13
Provide a way to identify operations that should not be retried

MattBrittan:
This is an initial solution to issue #214 (the autopaho queue would loop infinitely attempting to resend invalid messages). Further work is needed, but this deals with the main issue.

Ref #214

### DIFF
--- a/autopaho/auto.go
+++ b/autopaho/auto.go
@@ -594,7 +594,12 @@ connectionLoop:
 				c.debug.Printf("publishing message from queue with topic %s", pub2.Topic)
 				if _, err = cli.PublishWithOptions(ctx, &pub2, paho.PublishOptions{Method: paho.PublishMethod_AsyncSend}); err != nil {
 					c.errors.Printf("error publishing from queue: %s", err)
-					if errors.Is(err, paho.ErrNetworkErrorAfterStored) { // Message in session so remove from queue
+					if errors.Is(err, paho.ErrInvalidArguments) { // Some errors should not be retried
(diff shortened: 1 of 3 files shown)


autopaho/auto.go
```diff
@@ -594,7 +594,12 @@ connectionLoop:
 				c.debug.Printf("publishing message from queue with topic %s", pub2.Topic)
 				if _, err = cli.PublishWithOptions(ctx, &pub2, paho.PublishOptions{Method: paho.PublishMethod_AsyncSend}); err != nil {
 					c.errors.Printf("error publishing from queue: %s", err)
-					if errors.Is(err, paho.ErrNetworkErrorAfterStored) { // Message in session so remove from queue
+					if errors.Is(err, paho.ErrInvalidArguments) { // Some errors should not be retried
+						if err := entry.Remove(); err != nil {
+							c.errors.Printf("error removing queue entry: %s", err)
+						}
+						// Need a way to notify the user of this
```
vishnureddy17, line 601: I think having a callback for `PublishViaQueue` would be useful so the user can be notified. That same callback could be passed to the call to `PublishWithOptions` if/when callbacks are added for async published in `paho`

vishnureddy17 (approved):

autopaho/auto.go
MattBrittan: Agreed - but I think that can be added at a later date (I think this should be done as part of #216; the idea being that the callback should be called whenever the publish is complete whether successfully or stopped due to error).

PR #228, opened 2024-01-14
Reset DefaultPinger to reconnect to server

### LINKED ISSUE #227: Autopaho set DefaultPinger on ClientConfig can't reconnect to server after disconnection
**Describe the bug**
Autopaho set DefaultPinger on ClientConfig can't reconnect to the server after disconnection
Found this bug while using DefalutPinger with logger to test Pinger with autopaho (I guess this is only usecase for now)

**To reproduce**
1. Set PingHandler to DefaultPinger on ClientConfig and autopaho.NewConnection
```go
pinger := paho.NewDefaultPinger()
cfg := autopaho.ClientConfig{
    ...
        ....
        PingHandler: pinger,
        ....
    }
    ...
}
cm, err := autopaho.NewConnection(ctx, cfg)
```
2. Disconnect client network and wait for client being closed
3. Reconnect network -> But autopaho can't reconnect to the server

**Debug output**
```
2024/01/13 10:00:54 connecting
2024/01/13 10:00:54 sending CONNECT
2024/01/13 10:00:54 waiting for CONNACK/AUTH
2024/01/13 10:00:55 received CONNACK
mqtt connection up
client error: ping handler error: Run() already called
2024/01/13 10:00:55 received CONNACK, starting PingHandler
2024/01/13 10:00:55 starting publish packets loop
2024/01/13 10:00:55 starting incoming
2024/01/13 10:00:55 subscribing to …
2024/01/13 10:00:55 PacketSent() returning due to done channel
2024/01/13 10:00:55 waiting for SUBACK
2024/01/13 10:00:55 returning from ping handler worker
2024/01/13 10:00:55 error called: ping handler error: Run() already called
2024/01/13 10:00:55 client stopped
2024/01/13 10:00:55 stop() called with error: <nil>
2024/01/13 10:00:55 ping stopped
2024/01/13 10:00:55 conn closed
2024/01/13 10:00:55 acks tracker reset
2024/01/13 10:00:55 session updated, waiting on workers
2024/01/13 10:00:55 client stopping, incoming stopping
2024/01/13 10:00:55 returning from incoming worker
2024/01/13 10:00:55 queue got connection
2024/01/13 10:00:55 returning from publish packets loop worker
2024/01/13 10:00:55 workers done
2024/01/13 10:00:55 handleError received error: ping handler error: Run() already called
2024/01/13 10:00:55 handleError passed error on: ping handler error: Run() already called
```

**Expected behaviour**
autopaho should be able to reconnect to the server even previous client was closed.

**Software used:**
- Client version: e379c50417faa1e52adef46ea24f25737a0df707

**Additional context**
It happens because the values of Pinger in paho.ClientConfig is changed and autopaho uses this changed values to reconnect to the server
```go
func (p *DefaultPinger) Run(conn net.Conn, keepAlive uint16) error {
	if keepAlive == 0 {
		p.debug.Println("Run() returning immediately due to keepAlive == 0")
		return nil
	}
	if conn == nil {
		return fmt.Errorf("conn is nil")
	}
	p.mu.Lock()
	if p.timer != nil {
		p.mu.Unlock()
		return fmt.Errorf("Run() already called")
	}
	select {
	case <-p.done:
		p.mu.Unlock()
		return fmt.Errorf("Run() called after stop()")
	default:
	}
	p.keepAlive = keepAlive
	p.conn = conn
	p.timer = time.AfterFunc(0, p.sendPingreq) // Immediately send first pingreq
	p.mu.Unlock()

	return <-p.errChan
}
```
timer is set while Run()
so the timer of the Pinger in paho.ClientConfig is not nil after that, and it causes `return fmt.Errorf("Run() already called")` while trying reconnection
It's not only problem of timer, but done(+ other channels I guess?) and stopOnce should also be initialized in order to work correctly

This is not a critical bug because users don't need logger for production, but it would be good to fix I guess.

minyukim:
closes #227

### DIFF
--- a/paho/client.go
+++ b/paho/client.go
@@ -48,6 +48,14 @@ var (
 	ErrInvalidArguments = errors.New("invalid argument") // If included (errors.Join) in an error, there is a problem with the arguments passed. Retrying on the same connection with the same arguments will not succeed.
 )
 
+type pingerError struct {
+	error
+}
+
+func (pe *pingerError) Error() string {
+	return pe.error.Error()
+}
+
 type (
 	PublishReceived struct {
 		Packet *Publish
@@ -348,7 +356,7 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
 		defer c.workers.Done()
 		defer c.debug.Println("returning from ping handler worker")
 		if err := c.config.PingHandler.Run(c.config.Conn, keepalive); err != nil {
-			go c.error(fmt.Errorf("ping handler error: %w", err))
+			go c.error(&pingerError{fmt.Errorf("ping handler error: %w", err)})
 		}
 	}()
 
@@ -548,7 +556,7 @@ func (c *Client) incoming() {
 	}
 }
 
-func (c *Client) close() {
+func (c *Client) close(e error) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
@@ -563,8 +571,10 @@ func (c *Client) close() {
 	close(c.stop)
 
 	c.debug.Println("client stopped")
-	c.config.PingHandler.Stop()
-	c.debug.Println("ping stopped")
+	if _, ok := e.(*pingerError); !ok {
+		c.config.PingHandler.Stop()
+		c.debug.Println("ping stopped")
+	}
 	_ = c.config.Conn.Close()
 	c.debug.Println("conn closed")
 	c.acksTracker.reset()
@@ -587,12 +597,12 @@ func (c *Client) close() {
 // It also closes the client network connection.
 func (c *Client) error(e error) {
 	c.debug.Println("error called:", e)
-	c.close()
+	c.close(e)
 	go c.config.OnClientError(e)
 }
 
 func (c *Client) serverDisconnect(d *Disconnect) {
-	c.close()
+	c.close(nil)
 	c.debug.Println("calling OnServerDisconnect")
 	go c.config.OnServerDisconnect(d)
 }
@@ -973,7 +983,7 @@ func (c *Client) Disconnect(d *Disconnect) error {
 	c.debug.Println("disconnecting", d)
 	_, err := d.Packet().WriteTo(c.config.Conn)
 
-	c.close()
+	c.close(nil)
 
 	return err
 }
--- a/paho/client_test.go
+++ b/paho/client_test.go
@@ -81,7 +81,7 @@ func TestClientConnect(t *testing.T) {
 		Conn: ts.ClientConn(),
 	})
 	require.NotNil(t, c)
-	defer c.close()
+	defer c.close(nil)
 	c.SetDebugLogger(clientLogger)
 
 	cp := &Connect{
@@ -119,7 +119,7 @@ func TestClientSubscribe(t *testing.T) {
 		Conn: ts.ClientConn(),
 	})
 	require.NotNil(t, c)
-	defer c.close()
+	defer c.close(nil)
 	c.SetDebugLogger(clientLogger)
 
 	c.stop = make(chan struct{})
@@ -164,7 +164,7 @@ func TestClientUnsubscribe(t *testing.T) {
 		Conn: ts.ClientConn(),
 	})
 	require.NotNil(t, c)
-	defer c.close()
+	defer c.close(nil)
 	c.SetDebugLogger(clientLogger)
 
 	c.stop = make(chan struct{})
@@ -202,7 +202,7 @@ func TestClientPublishQoS0(t *testing.T) {
 		Conn: ts.ClientConn(),
 	})
 	require.NotNil(t, c)
-	defer c.close()
+	defer c.close(nil)
 	c.SetDebugLogger(clientLogger)
 
 	c.stop = make(chan struct{})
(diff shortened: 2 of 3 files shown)


paho/client.go
```diff
@@ -563,8 +571,10 @@ func (c *Client) close() {
 	close(c.stop)
 
 	c.debug.Println("client stopped")
-	c.config.PingHandler.Stop()
-	c.debug.Println("ping stopped")
+	if _, ok := e.(*pingerError); !ok {
```
MattBrittan, line 574: Consider what could happen if the connection drops when a ping is scheduled:

1. `c.error` called by something else
2. `pinger` exits with "failed to send PINGREQ" (having called `reset`)
3. `c.error` call from step 1 runs and stops the refreshed pinger (meaning that all future calls to `Run` will fail).

I realise that the above sequence of events is pretty unlikely but I believe it's possible (and it would be pretty hard to trace!). I think it's preferable for the `reset` to happen in `Run` meaning `Run` will work for a new `DefaultPinger` or following a clean shutdown (the user must always wait for `Run` to terminate before calling it again). This should be documented in the `interface` to make it clear that the `pinger` is reusable.

Widgle: Yes I think that scenario is possible too. I'll think about it more.

Widgle: @MattBrittan What about adding `Reset()` to Pinger interface like this
```go
type Pinger interface {
	// Run() starts the pinger. It blocks until the pinger is stopped.
	// If the pinger stops due to an error, it returns the error.
	// If the keepAlive is 0, it returns nil immediately.
	// Run() must be called only once.
	Run(conn net.Conn, keepAlive uint16) error

	// Stop() gracefully stops the pinger.
	Stop()

	// Reset() resets the pinger to be reusable.
	Reset()

	// PacketSent() is called when a packet is sent to the server.
	PacketSent()

	// PingResp() is called when a PINGRESP is received from the server.
	PingResp()

	// SetDebug() sets the logger for debugging.
	// It is not thread-safe and must be called before Run() to avoid race conditions.
	SetDebug(log.Logger)
}
```
and call it right after closing client?
```go
func (c *Client) close() {
	c.mu.Lock()
	defer c.mu.Unlock()

	defer c.config.PingHandler.Reset()

	select {
	case <-c.stop:
		// already shutting down, return when shutdown complete
		<-c.done
		return
	default:
	}

	close(c.stop)

	c.debug.Println("client stopped")
	c.config.PingHandler.Stop()
	c.debug.Println("ping stopped")
	_ = c.config.Conn.Close()
	c.debug.Println("conn closed")
	c.acksTracker.reset()
	c.debug.Println("acks tracker reset")
	c.config.Session.ConnectionLost(nil)
	if c.config.autoCloseSession {
		if err := c.config.Session.Close(); err != nil {
			c.errors.Println("error closing session", err)
		}
	}
	c.debug.Println("session updated, waiting on workers")
	c.workers.Wait()
	c.debug.Println("workers done")
	close(c.done)
}
```

MattBrittan: The problem I see with this is "what should the Pinger do if `Reset()` is called whilst the pinger is running". We may just say "this can't happen" (probably a mistake!) but then there is no real benefit to having a separate function (you can effectively call Reset from within `Run`). If we agree that `Run` could conceivably be called before `Stop` completes then the question becomes "what can you do about it" (only think I can come up with is to stop the old one and log a message).

As such I don't think there is much value in adding `Reset` and it's probably simplest/safest if  `Run`:

- Checks if another `Run` is active and, if so:
   - Stop it (`Run` should return an error so it will be logged etc). `Run` may need to wait for `Run` to terminate.
- Reset things
- Start the new process

MattBrittan: Sorry - ran out of time today so will try to have a look tomorrow.

Widgle: @MattBrittan No problem.
Are we 100% sure that `Run()` will always be called before `Stop()`? 
In this scenario,
1. `c.close()` is called right after `c.Connect()`
2. `Stop()` is called before calling `Run()` in gorutine

I think it can be possible because we don't wait for calling `Run()` in `c.Connect()`, although it will always never happen.
To be 100% sure, then I guess we need to separate `Run()` to things like `Start()` and `Wait()` like below, so that `c.Connect()` can wait for `Start()` of PingHandler.

```go
// client.go
func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
	...
        pingerWait := make(chan struct{})
	c.debug.Println("received CONNACK, starting PingHandler")
	c.workers.Add(1)
	go func() {
		defer c.workers.Done()
		defer c.debug.Println("returning from ping handler worker")
		if err := c.config.PingHandler.Start(c.config.Conn, keepalive); err != nil {
			...
		}
		close(pingerWait)
		if err:= c.config.PingHandler.Wait(); err != nil {
			go c.error(fmt.Errorf("ping handler error: %w", err))
		}
	}()
	...

	<- pingerWait

	return ca, nil	
}

// pinger.go
func (p *DefaultPinger) Start(conn net.Conn, keepAlive uint16) error {
	...	
}
func (p *DefaultPinger) Wait() error {
	return <-p.errChan
}
```

MattBrittan: Yep, I had been wondering if it would make more sense to pass `Run` a `Context` and use that for termination (was going to mock this up today but ran out of time).

minyukim:
PR https://github.com/eclipse/paho.golang/pull/229 dealt with the same issue, so close this.

Issue #148, opened 2023-07-27
What is Client.mu protecting?


vishnureddy17:
In `client.go`, there is a mutex `c.mu`, but it is unclear to me what exactly it's protecting. 

https://github.com/eclipse/paho.golang/blob/6c9185827625a60bd3c139d6ff63fce8846979ed/paho/client.go#L71

MattBrittan:
Looks like it's mainly protecting `c.raCtx` (most obviously) and the channels; for instance:

```
c.mu.Lock()
select {
case <-c.stop:
	c.mu.Unlock()
	return
default:
	c.publishPackets <- pb
	c.mu.Unlock()
}
```

It's use here prevents `Close` from running before `c.publishPackets <- pb` completes. It would probably be possible to simplify this but avoiding deadlocks can be tricky!.

vishnureddy17:
Yep, I see it, thanks!

A potential improvement it so figures out how to remove this. Keeping the issue open for that reason (if you'd rather it be closed, feel free to close it)

MattBrittan:
I don't think we can remove it but it would be worth improving the documentation (it really should be locked before any property of `Client` is changed).

vishnureddy17:
Going to go ahead and close this issue, I don't think it's needed anymore.

PR #229, opened 2024-01-15
Tidy up Mu, termination process and keep alive

### LINKED ISSUE #227: Autopaho set DefaultPinger on ClientConfig can't reconnect to server after disconnection

MattBrittan:
* It was unclear what Client.mu protected.
* Using and using contexts to manage shutdown is easier to follow. 
* Reimplement Pinger using Context

Closes #227
Ref #148

### DIFF
--- a/paho/client.go
+++ b/paho/client.go
@@ -72,8 +72,9 @@ type (
 		Session          session.SessionManager
 		autoCloseSession bool
 
-		AuthHandler Auther
-		PingHandler Pinger
+		AuthHandler   Auther
+		PingHandler   Pinger
+		defaultPinger bool
 
 		// Router - new inbound messages will be passed to the `Route(*packets.Publish)` function.
 		//
@@ -207,6 +208,7 @@ func NewClient(conf ClientConfig) *Client {
 
 	if c.config.PingHandler == nil {
 		c.config.PingHandler = NewDefaultPinger()
+		c.config.defaultPinger = true
 	}
 	if c.config.OnClientError == nil {
 		c.config.OnClientError = func(e error) {}
@@ -238,7 +240,8 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
 	c.connectCalled = true
 	c.connectCalledMu.Unlock()
 
-	// The passed in ctx applies to the connection process only. clientCtx applies to Client (signals
+	// The passed in ctx applies to the connection process only. clientCtx applies to Client (signals that the
+	// client should shut down).
 	clientCtx, cancelFunc := context.WithCancel(context.Background())
 	done := make(chan struct{})
 	cleanup := func() {
@@ -359,6 +362,9 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
 	}
 
 	c.debug.Println("received CONNACK, starting PingHandler")
+	if c.config.defaultPinger { // Debug logger is set after the client is created so need to copy it to pinger
+		c.config.PingHandler.SetDebug(c.debug)
+	}
 	c.workers.Add(1)
 	go func() {
 		defer c.workers.Done()
--- a/paho/client.go
+++ b/paho/client.go
@@ -72,9 +72,8 @@ type (
 		Session          session.SessionManager
 		autoCloseSession bool
 
-		AuthHandler   Auther
-		PingHandler   Pinger
-		defaultPinger bool
+		AuthHandler Auther
+		PingHandler Pinger
 
 		// Router - new inbound messages will be passed to the `Route(*packets.Publish)` function.
 		//
@@ -112,7 +111,6 @@ type (
 	}
 	// Client is the struct representing an MQTT client
 	Client struct {
-		mu     sync.Mutex
 		config ClientConfig
 
 		// OnPublishReceived copy of OnPublishReceived from ClientConfig (perhaps with added callback form Router)
@@ -120,10 +118,16 @@ type (
 		onPublishReceivedTracker []int // Used to track positions in above
 		onPublishReceivedMu      sync.Mutex
 
-		// authResponse is used for handling the MQTTv5 authentication exchange.
+		// authResponse is used for handling the MQTTv5 authentication exchange (MUST be buffered)
 		authResponse   chan<- packets.ControlPacket
-		stop           chan struct{}
-		done           chan struct{} // closed when shutdown complete (only valid after Connect returns nil error)
+		authResponseMu sync.Mutex // protects the above
+
+		cancelFunc func()
+
+		connectCalled   bool       // if true `Connect` has been called and a connection is being managed
+		connectCalledMu sync.Mutex // protects the above
+
+		done           <-chan struct{} // closed when shutdown complete (only valid after Connect returns nil error)
 		publishPackets chan *packets.Publish
 		acksTracker    acksTracker
 		workers        sync.WaitGroup
@@ -202,7 +206,6 @@ func NewClient(conf ClientConfig) *Client {
 	c.onPublishReceivedTracker = make([]int, len(c.onPublishReceived)) // Must have the same number of elements as onPublishReceived
 
 	if c.config.PingHandler == nil {
-		c.config.defaultPinger = true
 		c.config.PingHandler = NewDefaultPinger()
 	}
 	if c.config.OnClientError == nil {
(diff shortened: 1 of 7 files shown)


paho/pinger.go
```diff
@@ -81,106 +73,67 @@ func (p *DefaultPinger) Run(conn net.Conn, keepAlive uint16) error {
...
-		return fmt.Errorf("Run() already called")
+		return fmt.Errorf("Run() already in progress")
 	}
-	select {
-	case <-p.done:
+	p.running = true
+	p.mu.Unlock()
+	defer func() {
+		p.mu.Lock()
 		p.mu.Unlock()
```
Widgle, line 84: Guess defer is missing?

paho/pinger.go
```diff
@@ -81,106 +73,67 @@ func (p *DefaultPinger) Run(conn net.Conn, keepAlive uint16) error {
...
+			p.mu.Unlock()
+
+			if !lastPingSent.IsZero() && lastPingSent.After(lastPingResponse) {
+				p.debug.Printf("DefaultPinger PINGRESP timeout")
+				return fmt.Errorf("PINGRESP timed out")
+			}
+
+			if t.Before(pingDue) {
+				// A Control Packet has been sent since we last checked, meaning the ping can be delayed
+				timer.Reset(t.Sub(pingDue))
```
Widgle, line 109: Supposed to be pingDue.Sub(t)?

paho/client.go
```diff
@@ -224,17 +227,29 @@ func (c *Client) Connect(ctx context.Context, cp *Connect) (*Connack, error) {
...
+	// Connect twice is invalid.
+	c.connectCalledMu.Lock()
+	if c.connectCalled {
+		c.connectCalledMu.Unlock()
+		return nil, fmt.Errorf("connect must only be called once")
+	}
+	c.connectCalled = true
+	c.connectCalledMu.Unlock()
+
+	// The passed in ctx applies to the connection process only. clientCtx applies to Client (signals
```
vishnureddy17, line 241: Looks like you meant to say more here in this comment

paho/pinger.go
MattBrittan: OOps, corrected this

MattBrittan: Thanks, have moved the Unlock to the end.

paho/pinger.go
MattBrittan: Thanks, have moved the Unlock to the end.

paho/client.go
```diff
@@ -72,9 +72,8 @@ type (
 		Session          session.SessionManager
 		autoCloseSession bool
 
-		AuthHandler   Auther
-		PingHandler   Pinger
-		defaultPinger bool
```
vishnureddy17, line 77: This should be kept and its value should be used to determine whether `SetDebugLogger` should call `SetDebug` on the `Pinger`.

paho/client.go
MattBrittan: Thanks - completed this

paho/client.go
MattBrittan: Thanks - I guessed it was there for something like that. Added code to set the logger.

izarraga:
This bug has also happened to me.  When you commit PR#229 i will check again.

paho/pinger.go
```diff
@@ -81,106 +73,67 @@ func (p *DefaultPinger) Run(conn net.Conn, keepAlive uint16) error {
...
+				return fmt.Errorf("PINGRESP timed out")
+			}
+
+			if t.Before(pingDue) {
+				// A Control Packet has been sent since we last checked, meaning the ping can be delayed
+				timer.Reset(pingDue.Sub(t))
+				continue
+			}
+
+			if _, err := packets.NewControlPacket(packets.PINGREQ).WriteTo(conn); err != nil {
```
Lorderot, line 113: can this potentially block for a very long time, breaking the keep alive mechanism?

paho/pinger.go
```diff
@@ -81,106 +73,67 @@ func (p *DefaultPinger) Run(conn net.Conn, keepAlive uint16) error {
...
+	var lastPingSent time.Time
+	for {
+		select {
+		case <-ctx.Done():
+			timer.Stop() // We don't care if the timer has fired
+			return nil
+		case t := <-timer.C:
+			p.mu.Lock()
+			lastPingResponse := p.lastPingResponse
+			pingDue := p.lastPacketSent.Add(interval)
```
Lorderot, line 99: "packet is sent" isn't equal to "packet is received". Are we sure that it always guarantees that the connection is not broken? Do we always receive something from the cloud in response (e.g. publishes with QoS 0)? Looks like it can significantly postpone the closure of connection in some edge cases.

MattBrittan: This is as per MQTT-3.1.2-20 (the only requirement to send a keep alive is if no control packet has been sent within the time frame). If you have a really broken TCP connection (that is something is ACKing packets that have not actually been delivered) then this could be an issue (but the MQTT protocol relies on a standards compliant connection).

paho/pinger.go
Lorderot: If the connection is mostly idle, we rely on keepalives to detect network issues. PingHandler sends a PINGREQ, and if it doesn't receive a response within KeepAlive period, the client will be closed. However, this logic breaks in the edge cases when PingHandler hangs on the WriteTo(conn) operation due to some network issues; the call will be eventually unblocked due to the TCP timeout, but it could be way longer than the specified KeppAlive period. 

Isn't it better to send a packet in a separate goroutine; and in case a response is not received within the KeepAlive period, close the connection (which would unblock the potentially stuck goroutine)?

MattBrittan: >Isn't it better to send a packet in a separate goroutine

OK - when you said it could block I thought you meant it was blocked because a message was being transmitted. This can happen, if you send a large message over a slow link there is a potential for it to take longer than  `keepAlive` to be transmitted (and there is not much we can do about that).

I believe the defaults for TCP retransmission periods are something like 13 to 30 minutes on Linux and around 65 seconds on windows. so there does appear to be a potential for the  `Write` to block longer than the keepalive period.

So there may be a rationale for moving the `Send` to a goroutine but I think that should be a separate issue/PR (this PR was more about tidying things up and removing a potential deadlock so the way the packet is sent was not changed from previous code).

I would note that the much older `paho.mqtt.golang` sends pings in the same way (does not include any checks re how long the send takes) and I don't believe any issues have been raised.

Lorderot: Created a [PR](https://github.com/eclipse/paho.golang/pull/257).

Issue #233, opened 2024-01-29
Autopaho managePublishQueue() does not call Remove, Quarantine, or Leave if Reader() fails


vishnureddy17:
In autopaho, if `entry.Reader()` fails we do not call `Remove`, `Quarantine`, or `Leave` on the entry as required. I'm guessing there should be a call to `Leave()` in the case where `Reader()` fails. Happy to open a PR for this.
https://github.com/eclipse/paho.golang/blob/ef0065fea247d5fb388fa82390cedce694e844e1/autopaho/auto.go#L553-L562

MattBrittan:
Well spotted; happy to accept a PR. I'm not not sure if it's best to leave or Remove in this case; if we cannot read the message then what are the chances that this would succeed if we try again.

vishnureddy17:
> if we cannot read the message then what are the chances that this would succeed if we try again.

This is the case that prompted me to open #234

MattBrittan:
Just noticed a further bug here; we check `if errors.Is(err, queue.ErrEmpty) {` but ignore other errors (not sure what the right thing to do is; perhaps a small timeout then loop).

vishnureddy17:
> (not sure what the right thing to do is; perhaps a small timeout then loop).

That seems reasonable. I also think we should have that same timeout in the `default` case if `PublishWithOptions` fails with an error that is not `paho.ErrNetworkErrorAfterStored` or `paho.ErrInvalidArguments`. Right now, we immediately loop without a delay:
https://github.com/eclipse/paho.golang/blob/ef0065fea247d5fb388fa82390cedce694e844e1/autopaho/auto.go#L612-L621

Issue #232, opened 2024-01-29
PublishWithOptions() will not return ErrInvalidArguments when QoS is invalid


vishnureddy17:
https://github.com/eclipse/paho.golang/blob/ef0065fea247d5fb388fa82390cedce694e844e1/paho/client.go#L885

This should have `ErrInvalidArguments` in its tree so the application knows it is a fatal error. This is a very quick fix and I am happy to open a PR for it.

MattBrittan:
>This is a very quick fix and I am happy to open a PR for it.

Please do - keen to get a release out this week and would be nice to have this fixed.

vishnureddy17:
Will open a PR today or tomorrow (US Eastern time). Would you like me to incorporate a fix to https://github.com/eclipse/paho.golang/issues/233 as well? @MattBrittan

MattBrittan:
>Would you like me to incorporate a fix to https://github.com/eclipse/paho.golang/issues/233 as well

Yes please.

PR #235, opened 2024-01-29
address issues with queue error handling in autopaho, return ErrInvalidArguments for invalid publish QoS in paho

vishnureddy17:
close https://github.com/eclipse/paho.golang/issues/233
close https://github.com/eclipse/paho.golang/issues/232

### DIFF
--- a/autopaho/auto.go
+++ b/autopaho/auto.go
@@ -554,10 +554,20 @@ connectionLoop:
 				if errors.Is(err, queue.ErrEmpty) {
 					c.debug.Println("everything in queue transmitted")
 					continue queueLoop
+				} else if err != nil {
+					// if Peek() keeps returning errors, we will loop forever.
+					// see https://github.com/eclipse/paho.golang/issues/234
+					c.errors.Printf("error retrieving queue entry: %s", err)
+					time.Sleep(1 * time.Second)
+					continue
 				}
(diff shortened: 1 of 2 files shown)


vishnureddy17:
I noticed that the Pinger tests are failing, but they were failing before the changes here. Otherwise, the existing paho and autopaho tests pass.

autopaho/auto.go
```diff
@@ -554,10 +554,20 @@ connectionLoop:
 				if errors.Is(err, queue.ErrEmpty) {
 					c.debug.Println("everything in queue transmitted")
 					continue queueLoop
+				} else if err != nil {
+					// if Peek() keeps returning errors, we will loop forever.
+					// see https://github.com/eclipse/paho.golang/issues/234
+					c.errors.Printf("error retrieving queue entry: %s", err)
+					time.Sleep(1 * time.Second)
```
vishnureddy17, line 561: I've put a hardcoded 1 second delay any time we are retrying the same publish. Open to feedback on a different approach.

MattBrittan:
> I noticed that the Pinger tests are failing, but they were failing before the changes here. Otherwise, the existing paho and autopaho tests pass.

Tests all pass for me; I'm guessing there is a timing issue somewhere. Can you please raise an issue with details of the failing test?

PR #240, opened 2024-02-16
Remove MaximumQoS property from CONNECT packet

### LINKED ISSUE #161: ConnectProperties in cp_connect.go has a MaximumQOS field that is not part of the MQTT 5 specification
The MQTT 5 spec doesn't include a MaximumQOS field in CONNECT properties, but it is included as a field.

https://github.com/eclipse/paho.golang/blob/21b4033e32237c844f59fab2a2d90954c01c1a2d/paho/cp_connect.go#L29

tomatod:
Closes #161

According to [MQTTv5 spec](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901046), Maximum QoS property should not be included in CONNECT packet, but it may be included only in CONNACK packet.

Actually, I included that property in CONNECT and sent it to Mosquitto MQTT server, the connection was refused. 

So, I also think we should remove Maximum QoS property from CONNECT packet in this library.

If you like this, I hope you can review this request.

Best Regards,

note:
I couldn't pass the test by 'make test' command each in this branch and also master with the following message. It seems to not be related to the change of this branch. 

```
queue_test.go:208: timeout awaiting messages
...
FAIL    github.com/eclipse/paho.golang/autopaho 5.632s
```


tomatod:
I'll cancel this request and check the cause of error after all.

Issue #242, opened 2024-02-17
An autopaho test TestQueuedMessage may be not passed.


tomatod:
**Describe the bug**
I think this isn't a bug, but it could be reproduced in partical environments especially with low machine resources. So, as a attention, I reported this.

autopaho's main loop of connection manager is terminated in one second at TestQueuedMessage. The part of implemention is the following.
```
func TestQueuedMessages(t *testing.T) {
...
        // [tomatod's comment] longerDelay is one second
	ctx, cancel := context.WithTimeout(context.Background(), longerDelay)
	defer cancel()
	cm, err := NewConnection(ctx, config)
...
}
```
So, if your machine couldn't finish the test in one secode, the connection manager would be terminated and test would be failed in result.

**To reproduce**
- Environment: Memory: 6GiB / CPU: Intel(R) Pentium(R) Silver N5000 CPU @ 1.10GHz (4 core)
- Command: ```go test -coverprofile /tmp/autopaho_coverage.out -race -tags=unittest ./autopaho/ -v -count 1 -run TestQueuedMessages```

(I often use this low spec machine outside...)

**Debug output**
Hire is an example result of the above test command.
```
...
    test.go:53: 2024-02-17T16:54:18.467392298+09:00 test: innerCtx Done
    test.go:53: 2024-02-17T16:54:18.467846545+09:00 test: disconnecting &{<nil> 0}
    test.go:53: 2024-02-17T16:54:18.468537774+09:00 test: client stopping, incoming stopping
    test.go:53: 2024-02-17T16:54:18.468715317+09:00 test: returning from incoming worker
    test.go:53: 2024-02-17T16:54:18.46899835+09:00 test: returning from publish packets loop worker
    test.go:53: 2024-02-17T16:54:18.469265094+09:00 test: error called: EOF
    test.go:53: 2024-02-17T16:54:18.469104492+09:00 testServer: handleIncoming closed with error handleIncoming:ReadPacket: read tcp 127.0.0.1:46611->127.0.0.1:37874: use of closed network connection Remaining data: []
    test.go:53: 2024-02-17T16:54:18.469471831+09:00 test: returning from ping handler worker
    test.go:63: 2024-02-17T16:54:18.469598315+09:00 testServer:error closing ourConn: close tcp 127.0.0.1:46611->127.0.0.1:37874: use of closed network connection
    test.go:53: 2024-02-17T16:54:18.470503258+09:00 testServer: disconnected
[...]

PR #243, opened 2024-02-18
Remove MaximumQoS property from CONNECT packet #161

### LINKED ISSUE #161: ConnectProperties in cp_connect.go has a MaximumQOS field that is not part of the MQTT 5 specification

tomatod:
**Do not create a Pull Request without creating an issue first**
#161 was created related to this request.

**The design of your solution to the problem should be discussed and agreed in an issue before submitting the PR**
According to [MQTTv5 spec](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901046), Maximum QoS property should not be included in CONNECT packet. Actually, when I tried to send CONNECT packet included that property to Mosquitto MQTT server, the connection was refused by Mosquitto. So, I think we should remove Maximum QoS property from CONNECT packet in this library.

note: This request is same with #240, but I remaked the request after some confirmations, because there was a problem with test on my envrionment (#242). As a result, the problem wasn't probably related to this change.

had already done

**Testing**
- ```make test``` was passed. I sometimes catched an issue with #242, but tests were passed temporarily with making the timeout of autopaho's cm longer.
- I confirmed almost all parts related to this change with grep ''QOS" or "QoS" (including "MaximumQoS" and "MaximumQOS") just in case.

**Closing issues**
closes #161

### DIFF
--- a/packets/properties.go
+++ b/packets/properties.go
@@ -300,11 +300,6 @@ func (i *Properties) Pack(p byte) []byte {
 			writeUint16(*i.TopicAliasMaximum, &b)
 		}
 
-		if i.MaximumQOS != nil {
-			b.WriteByte(PropMaximumQOS)
-			b.WriteByte(*i.MaximumQOS)
-		}
-
 		if i.MaximumPacketSize != nil {
 			b.WriteByte(PropMaximumPacketSize)
 			writeUint32(*i.MaximumPacketSize, &b)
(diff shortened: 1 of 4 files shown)


MattBrittan:
Thanks very much for submitting this.

PR #248, opened 2024-04-10
fix: respect context cancellation while using the provided Dialer

Lorderot:
A ConnectTimeout setting must be respected while connecting through the proxy:
1. Call DialContext if the custom dialer implements it.
2. Perform the blocking Dial call in a new goroutine, and respect context cancellation (potentially "leaked" goroutine)


MattBrittan:
Thanks for the PR; looks good to me (as `proxy.Dial` calls `FromEnvironment()` I would not expect this to break any existing use-cases).

Note: Please don't chase these (especially after one day). We will look at PR's but it may sometimes take a while (having accepted, very small, PR's that broke things I want to have a decent look before accepting).

PR #252, opened 2024-04-19
Add backoff strategy interfaces and implementations

### LINKED ISSUE #249: Support exponential backoff for reconnects
When there is a large number of clients and there is some kind of synchronization event (e.g. broker restart, update, etc.) the reconnect will trigger a reconnect storm.

I'd like to be able to provide a "backoff strategy".
For mass-clients it could implement an exponential backoff, for some e.g. some kind of backend service it could use a fixed value similar to what we have right now.
Both strategies could be implemented as generic solutions which would allow user to customize then according to their needs / infrastructure abilities.

A rate limiter on the broker side sure helps but is more of an add-on than an alternative to a backof strategy.

**Additional context**
No additional context.

ViToni:
This PR introduces the ability to use a backoff strategy for reconnect attempts.

As default a backoff strategy is used whichs return the same backoff value mimicking the previous behaviour.
In addition there is an exponential backoff available which will increase the max value on each attempt up to a configurable limit.
The created backoff is a random value between a configured non-zero minimum and the increasing max (up to the configured limit). Using a random value from an increasing range helps spread client reconnect attempts.

**Testing**

Existing code has been changed to use the same duration as before but now uses the constant backoff implementation instead. All tests pass.

Additionally there are dedicated tests for:

- Constant backoff
- Exponential backoff

**Closing issues**

closes #249 `

### DIFF
--- a/autopaho/auto.go
+++ b/autopaho/auto.go
@@ -70,9 +70,11 @@ type ClientConfig struct {
 	CleanStartOnInitialConnection bool        //  Clean Start flag, if true, existing session information will be cleared on the first connection (it will be false for subsequent connections)
 	SessionExpiryInterval         uint32      // Session Expiry Interval in seconds (if 0 the Session ends when the Network Connection is closed)
 
-	ConnectRetryDelay time.Duration    // How long to wait between connection attempts (defaults to 10s)
-	ConnectTimeout    time.Duration    // How long to wait for the connection process to complete (defaults to 10s)
-	WebSocketCfg      *WebSocketConfig // Enables customisation of the websocket connection
+	// Deprecated: ConnectRetryDelay is deprecated and its functionality is replaced by ReconnectBackoffStrategy.
+	ConnectRetryDelay        time.Duration    // How long to wait between connection attempts (defaults to 10s)
+	ReconnectBackoffStrategy BackoffStrategy  // How long to wait between connection attempts (defaults to 10s)
+	ConnectTimeout           time.Duration    // How long to wait for the connection process to complete (defaults to 10s)
+	WebSocketCfg             *WebSocketConfig // Enables customisation of the websocket connection
 
 	Queue queue.Queue // Used to queue up publish messages (if nil an error will be returned if publish could not be transmitted)
 
@@ -242,8 +244,15 @@ func NewConnection(ctx context.Context, cfg ClientConfig) (*ConnectionManager, e
 	if cfg.Errors == nil {
 		cfg.Errors = log.NOOPLogger{}
 	}
-	if cfg.ConnectRetryDelay == 0 {
-		cfg.ConnectRetryDelay = 10 * time.Second
+	if cfg.ReconnectBackoffStrategy == nil {
+		// for backwards compatibility we check for ConnectRetryDelay first
+		// before using the default constant backoff strategy (which behaves
+		// identically to the previous behaviour)
+		if cfg.ConnectRetryDelay == 0 {
+			cfg.ReconnectBackoffStrategy = NewConstantBackoffStrategy(10 * time.Second)
+		} else {
+			cfg.ReconnectBackoffStrategy = NewConstantBackoffStrategy(cfg.ConnectRetryDelay)
+		}
 	}
 	if cfg.ConnectTimeout == 0 {
 		cfg.ConnectTimeout = 10 * time.Second
--- a/autopaho/backoff.go
+++ b/autopaho/backoff.go
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ *
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License v2.0
+ *  and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    https://www.eclipse.org/legal/epl-2.0/
+ *  and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ *  SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+package autopaho
+
+import (
+	"time"
+)
+
+// The BackoffStrategy is a container for the configuration of a given strategy
+// whereas the returned Backoff is the actual implementation of the strategy.
+//
+// As such BackoffStrategy instances are supposed to be thread safe and freely sharable between users.
+type BackoffStrategy interface {
+	// Returns a configured backoff instance.
+	Backoff() Backoff
+}
+
+// Backoff represents a configured instance of the strategy which computes the
+// next backoff duration.
+//
+// NOTE: Generally instances are NOT supposed to be reused or shared.
+type Backoff interface {
+	// Returns the next backoff duration.
+	Next() time.Duration
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// helper structs for backoff implementations
+////////////////////////////////////////////////////////////////////////////////
+
+// The BackoffDelegate implements the Backoff interface and delegates backoff
+// computation to the internal function property.
+type BackoffDelegate struct {
+	next func() time.Duration
+}
+
+func (b *BackoffDelegate) Next() time.Duration {
+	return b.next()
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// implementation for constant backoff
+////////////////////////////////////////////////////////////////////////////////
+
+// The ConstantBackoffStrategy implements the BackoffStrategy interface and
+// provides instances of constant duration.
+//
+// As the duration is constant, the strategy and the actual backoff are thread
+// safe and can be shared.
+type ConstantBackoffStrategy struct {
+	backoff Backoff
+}
+
+func (bs *ConstantBackoffStrategy) Backoff() Backoff {
+	return bs.backoff
+}
+
+// Creates a new ConstantBackoffStrategy with the specified delay.
+func NewConstantBackoffStrategy(delay time.Duration) *ConstantBackoffStrategy {
+	backoff := &BackoffDelegate{
+		next: func() time.Duration {
+			return delay
+		},
+	}
+
+	return &ConstantBackoffStrategy{
+		backoff: backoff,
+	}
+}
--- a/autopaho/backoff.go
+++ b/autopaho/backoff.go
@@ -16,6 +16,7 @@
 package autopaho
 
 import (
+	"math/rand"
 	"time"
 )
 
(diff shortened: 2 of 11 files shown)


autopaho/backoff.go
```diff
@@ -0,0 +1,193 @@
...
+package autopaho
+
+import (
+	"math/rand"
+	"time"
+)
+
+// The BackoffStrategy is a container for the configuration of a given strategy whereas the returned Backoff is the actual implementation
+// As such BackoffStrategy instances are supposed to be thread safe and freely sharable between users.
+type BackoffStrategy interface {
```
MattBrittan, line 25: I'm suggesting fairly major changes to handle an issue you were not aware of so feel free to tell me to get lost!

Consider the following possibility:
1. Establish connection and send CONNECT (so `establishServerConnection` has completed)
2. Broker rejects CONNECT (say bad password) so look to 1

This can result in us hammering the broker with CONNECT packets (not a good look!). It would be nice if we could solve this issue at the same time as dealing with the backoff on the network level connection.

I've had a quick think about options and wonder if the following might work (and provide a solution that's a bit simpler than your approach):

```
ConnectionBackoff func(ctx Context.Context, establishConnection func(ctx Context.Context) error) error // Passed function that will establish network connection and calls it until a `nil` error is returned or the context is cancelled.
```

Then in `establishServerConnection` you would do something like:

```
var cli *paho.Client,
var connack *paho.Connack)

err := cfg.ConnectionBackoff( ctx, func(ctx Context.Context) error {
     for _, u := range cfg.ServerUrls {
		connectionCtx, cancelConnCtx := context.WithTimeout(ctx, cfg.ConnectTimeout)
....
    }
}
if err != nil { // this would only happen when context cancelled

}
return cli, connack
```

The advantage of this is that it means that the `ConnectionBackoff ` has complete control of the process (including resetting, or not, between connections). `ConstantBackoff` could be very simple (pseudo code!):

```
type ConstantBackoff struct {
	delay       time.Duration
	lastAttempt time.Time
}

func (cb *ConstantBackoff) Run(ctx context.Context, establishConnection func(ctx Context.Context) error) {
	// The previous connection attempt may have been very recent; we want at least delay between attempts
	select {
	case time.After(cb.delay - time.Since(cb.lastAttempt)): // Might be negative; i.e. no delay
	case <-ctx.Done():
		return ctx.Err()
	}

	for {
		err := establishConnection(ctx)
		if err == nil || ctx.Err() != nil {
			return err
		}
		select {
		case <-time.After(cb.delay):
		case <-ctx.Done():
			return ctx.Err()
		}
	}
}
```

Other backoffs could be a bit more complex (and take into account whatever factors are relevant).

MattBrittan (commented):
Sorry - I'm suggesting a big change here to deal with an issue that is completly missed by the current `ConnectRetryDelay` . Please feel free to say "no"!

autopaho/auto.go
```diff
@@ -70,9 +70,9 @@ type ClientConfig struct {
 	CleanStartOnInitialConnection bool        //  Clean Start flag, if true, existing session information will be cleared on the first connection (it will be false for subsequent connections)
 	SessionExpiryInterval         uint32      // Session Expiry Interval in seconds (if 0 the Session ends when the Network Connection is closed)
 
-	ConnectRetryDelay time.Duration    // How long to wait between connection attempts (defaults to 10s)
-	ConnectTimeout    time.Duration    // How long to wait for the connection process to complete (defaults to 10s)
-	WebSocketCfg      *WebSocketConfig // Enables customisation of the websocket connection
+	ReconnectBackoffStrategy BackoffStrategy  // How long to wait between connection attempts (defaults to 10s)
```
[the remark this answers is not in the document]
ViToni: I'd prefer to have only one way to support backoffs to avoid double code. Using the new interface would mean a one-time effort on the consumer side. But it's your call.

autopaho/auto.go
```diff
@@ -242,8 +242,8 @@ func NewConnection(ctx context.Context, cfg ClientConfig) (*ConnectionManager, e
 	if cfg.Errors == nil {
 		cfg.Errors = log.NOOPLogger{}
 	}
-	if cfg.ConnectRetryDelay == 0 {
-		cfg.ConnectRetryDelay = 10 * time.Second
+	if cfg.ReconnectBackoffStrategy == nil {
+		cfg.ReconnectBackoffStrategy = NewConstantBackoffStrategy(10 * time.Second)
```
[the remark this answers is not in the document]
ViToni: Could be done. Depends one the comment above.

autopaho/backoff.go
ViToni: My approach was to have a simple implementation. Therefore one can re-use / share the `BackoffStrategy` while the actual `Backoff` is "throwaway". It doesn't need a "reset" of anything fancy.
In my experience these mechanisms, where the server could potentially say: "come back in one hour", sound nice but are not worth the hassle. Often the "one hour" is either to early or to late.
Having a rate limiter on the server side and an exponential backoff on the client side, has worked for me with planned and unplanned outages in the past. But that's only my use case... so your mileage might vary.
One aspect which 
Either way, IMO something like taking care of CONNECT should have a dedicated PR and discussion (as it is seems to me way more complex than the backoff itself).

autopaho/auto.go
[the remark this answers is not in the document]
MattBrittan: As we are still pre-V1 a breaking change like this is possible; however my preference would be to make as many breaking changes as possible in one release (there are a few options that have changed and are now marked as depreciated). I'm happy either way as I we are getting fairly close to V1, but if you do make the breaking change then I'll hold off making a release until the other breaking changes are made (think it's best to put as many of these in one release as possible so users don't need to update their code with every release).

autopaho/backoff.go
MattBrittan: > IMO something like taking care of CONNECT should have a dedicated PR and discussion (as it seems to me way more complex than the backoff itself).

Sorry - I made an error here (the issue with working on multiple clients concurrently!). In this client `establishServerConnection` only returns after the `CONNACK` is received. As such this is not the issue I thought it was (and your current strategy would work).

Note that the reason I raised this (and want to put some time into considering this change) is due to experience with the V3 client where we hit issues because the options became very complicated, so I'm careful with new options in this project (once an option is added it's very hard to remove!).

>where the server could potentially say: "come back in one hour",

The one thing that may be a little bit like the scenario you mention ("come back later") is duplicate client id connect loops. This is very common for users new to the protocol where they have two clients with the same client ID and end up with a connect loop (two connections "fighting"). With v3.1 this was difficult to detect (the server just dropped the connection) but with V5 a specific Disconnect would be sent (Reason Code 0x8E - Session taken over) so there is a potential that we could handle it in the future. I only mention this as being able to prevent connection storms when this is happening would be beneficial longer term (so would be nice to have the option of adding support for this).

>My approach was to have a simple implementation. Therefore one can re-use / share the BackoffStrategy while the actual Backoff is "throwaway". It doesn't need a "reset" of anything fancy.

I understand that; however part of my reaction here might have been because your code does look a bit Java'ish. I avoided this in the last set of comments because it's highly subjective; but as you specifically asked "So if I hit any no-Go-s" here are a few thoughts (I'll add a few more comments in the code):

* The structure made me think of the Java factory pattern; obviously this pattern is used in Go but it's a lot less common than in Java.
* Ive not seen use of underscore prefixes (i.e.  `_Duration`) other than in unexported globals. I'd expect this variable to be called `delay` or similar (with the lower case `d` indicating it's private).

My suggestion would be to simplify this to (using the `Reset` option you felt complicated things):

```
// backoff represents a configured instance of the strategy which computes the next backoff duration.
type backoff interface {
        Reset() 
	NextBackOff() time.Duration
}
```

There are a few reasons for this:

It's simple.

It's compatible with [`cenkalti/backoff`](https://github.com/cenkalti/backoff) which is the most well known back-off library I'm aware of (so would provide a range of options users could simply plug in). Note: I would prefer to avoid a dependency so the basic strategies would be implemented in this library (which you have done).

It allows authors to consider previous state (e.g. never attempt to connect twice in two minutes) which is doable, but more difficult, with the current structure.

We can extend it in the future without breaking users code. For example to handle the connection fight situation mentioned above we could add:

``` 
type backoffWithDisconnected interface {
     backoff 
    Disconnected(*package.Disconnect) // called when Disconnect packet received
}
```

Then, when a disconnect packet is received, we would do something like:

```
if on, dwd := cfg.Backoff.(backoffWithDisconnected) ; ok {
   dwd.Disconnected(pkt)
}
```

This is just an off the cuff thought; so treat as pseudo code but aims to show why this provided more flexibility.

Note: Once more I'm not saying that this is the right approach but do want to be careful with this because once an interface is selected I think we need to consider it locked in.

autopaho/backoff.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+}
+
+func (c *ConstantBackoff) Next() time.Duration {
+	return c._Duration
+}
+
+// The ConstantBackoffStrategy implements the BackoffStrategy interface and provides instances a constant duration.
+type ConstantBackoffStrategy struct {
+	BackoffStrategy
```
MattBrittan, line 60: Why embed `BackoffStrategy`? I suspect this may be a fallback to Java where you would need to declare what interface you are implementing but it's not needed here (and is harmful because it would prevent the compiler from warning you if your implementation of `GetBackoff()` has the wrong signature.

autopaho/backoff.go
```diff
@@ -0,0 +1,193 @@
...
+	Next() time.Duration
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// implementation for constant backoff
+////////////////////////////////////////////////////////////////////////////////
+
+// The ConstantBackoff implements the Backoff interface and provides a constant duration.
+type ConstantBackoff struct {
+	Backoff
```
MattBrittan, line 43: No need to embed the `interface` here. Doing this will prevent the compiler from identifying issues (and could lead to a runtime crash [see here](https://go.dev/play/p/8aLFQRHSK9j))

autopaho/backoff.go
```diff
@@ -0,0 +1,193 @@
...
+	_Duration time.Duration
```
MattBrittan, line 44: An underscore prefix is not a go standard; just use lower case for the first letter (looks a bit weird for a while but you get used to it).

MattBrittan (commented):
A few of those style/possible java influences I mentioned.

autopaho/backoff.go
ViToni: > I understand that; however part of my reaction here might have been because your code does look a bit Java'ish.

You are totally right. Funny how it shows. I've been doing Java for many years and as you correctly guessed this code originates from Java. In Java it was very simple:

```
public interface BackoffStrategy {
    Iterator<Duration> backoff();
}
```

As the custom `Iterator` was a nested class in the implementation it could easily/safely resuse final fields from the outer class.

But somehow I was not able to implement something similar simple in Go.

autopaho/backoff.go
ViToni: Many thanks for pointintg this out. I wasn't aware this would be actuall harmful. This was definitely my Java way of thinking...

autopaho/backoff.go
ViToni: Again thanks for pointing it out (and for the example code showing the issue in action)

ViToni:
@MattBrittan The PR now has been updated. Hopefully it looks more Go-ish now.
I'd be more than happy for any feedback.

autopaho/auto.go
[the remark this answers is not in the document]
ViToni: Fallback to `ConnectRetryDelay` is now in place.

autopaho/examples/backoff/backoff.go
```diff
@@ -0,0 +1,51 @@
+/*
```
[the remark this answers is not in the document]
ViToni: Actually it helped (me) visualize the range of the `ExponentialBackoffStrategy`. There was a tiny bug while refactoring, which would have been difficult to catch due to the randomness of values but way easily spotable using this file.

MattBrittan:
> @MattBrittan The PR now has been updated. Hopefully it looks more Go-ish now. I'd be more than happy for any feedback.

Thanks - I'll try to have a look next week (sorry, have limited time to spend on this currently).

ViToni:
Superseded by #258.

PR #255, opened 2024-05-11
PublishWithOptions QoS0 return empty PublishRespoonse

### LINKED ISSUE #159: autopaho publisher example: panic due to nil pointer
I get a nil pointer dereference in the [autopaho publisher example](https://github.com/eclipse/paho.golang/blob/21b4033e32237c844f59fab2a2d90954c01c1a2d/autopaho/examples/docker/publisher/main.go#L93) because `pr` can be nil even if err is also nil.

wic006:
PublishWithOptions may return both nil error and nil PublishResponse when QoS0 is used, which can potentially result in nil pointer exceptions. This commit updates PublishWithOptions to return a pointer to an empty PublishResponse when QoS0 is used to eliminate this risk.

Closes #159

### DIFF
--- a/paho/client.go
+++ b/paho/client.go
@@ -874,7 +874,7 @@ func (c *Client) PublishWithOptions(ctx context.Context, p *Publish, o PublishOp
 			return nil, err
 		}
 		c.config.PingHandler.PacketSent()
-		return nil, nil
+		return &PublishResponse{}, nil
 	case 1, 2:
 		return c.publishQoS12(ctx, pb, o)
 	}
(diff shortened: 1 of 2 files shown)


MattBrittan:
Thanks - I'm not hugely happy with this, but think it's probably the best way forward. Returning a `nil` pointer when there is no error is more likely to cause a runtime crash than returning a value (and the default value will have a status of `0` - `PubackSuccess` so seems reasonable). If you get a chance it would be good to have a comment that explains this (just because it's somewhat unexpected that there would be a response to a QOS0 message)..

PR #257, opened 2024-05-22
Improve default ping handler

Lorderot:
### Current state
In order to detect network issues in the application layer, PingHandler sends a PINGREQ, and if it doesn't receive a response within the specified KeepAlive period, the connection is closed. However, if the network is very slow and unstable, the WriteTo(conn) operation (which sends a PINGREQ) may block for quite some time, during which context cancellation and KeepAlive timeout are ignored. The call must be eventually unblocked due to TCP timeouts. (Note: I have an example where connection establishment through the proxy hung for 1 hour; it's already fixed with proxy.Dial and ContextDialer. However, I suspect the same could happen here as well)
### Proposal
Send a PINGREQ request in a separate goroutine. In case of KeepAlive timeout or context cancellation, the main goroutine does not wait until the write operation finishes. The spawned goroutine will not be leaked if the connection is closed (this must unblock the write operation).

### DIFF
--- a/paho/pinger.go
+++ b/paho/pinger.go
@@ -87,13 +87,14 @@ func (p *DefaultPinger) Run(ctx context.Context, conn net.Conn, keepAlive uint16
 
 	interval := time.Duration(keepAlive) * time.Second
 	timer := time.NewTimer(0) // Immediately send first pingreq
+	// If timer is not stopped, it cannot be garbage collected until it fires.
+	defer timer.Stop()
 	var lastPingSent time.Time
 	// errCh should be buffered, so that the goroutine sending the error does not block if the context is cancelled
 	errCh := make(chan error, 1)
 	for {
 		select {
 		case <-ctx.Done():
-			timer.Stop() // We don't care if the timer has fired
 			return nil
 		case t := <-timer.C:
 			p.mu.Lock()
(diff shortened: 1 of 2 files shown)


paho/pinger.go
```diff
@@ -109,13 +111,20 @@ func (p *DefaultPinger) Run(ctx context.Context, conn net.Conn, keepAlive uint16
...
-				return fmt.Errorf("failed to send PINGREQ: %w", err)
-			}
+			lastPingSent = time.Now()
+			go func() {
+				if _, err := packets.NewControlPacket(packets.PINGREQ).WriteTo(conn); err != nil {
+					p.debug.Printf("DefaultPinger packet write error: %v", err)
+					errCh <- fmt.Errorf("failed to send PINGREQ: %w", err)
+					return
+				}
+				errCh <- nil
```
MattBrittan, line 121: Why send a `nil` error? (it's ignored in the select)

paho/pinger.go
```diff
@@ -109,13 +111,20 @@ func (p *DefaultPinger) Run(ctx context.Context, conn net.Conn, keepAlive uint16
...
+			}()
 			timer.Reset(interval)
+		case err := <-errCh:
+			if err != nil {
+				return err
```
MattBrittan, line 126: Probably worth stopping the timer here too (in the previous version the `return` occurred before the timer had fired so this was not necessary). This is a minor point as the timer would fire eventually but think it's tidier (and there could be an issue in a connect/disconnect loop with a long keepalive period).

paho/pinger.go
```diff
@@ -109,13 +111,20 @@ func (p *DefaultPinger) Run(ctx context.Context, conn net.Conn, keepAlive uint16
...
 				continue
 			}
-
-			lastPingSent = time.Now() // set before sending because WriteTo may return after PINGRESP is handled
-			if _, err := packets.NewControlPacket(packets.PINGREQ).WriteTo(conn); err != nil {
-				p.debug.Printf("DefaultPinger packet write error: %v", err)
-				return fmt.Errorf("failed to send PINGREQ: %w", err)
-			}
+			lastPingSent = time.Now()
+			go func() {
```
MattBrittan, line 115: Think it would be work mentioning why this has been done and the potential issue (have seen this issue logged against other clients). I don't think the spec really covers this... e.g.
// WriteTo may not complete within keepalive period so send in goroutine. Note potential issue if huge message is sent over very slow link at the same time as keepalive (so Write is blocked by other message and keepalive is exceeded). This is seems fairly unlikely and there is no obvious way to avoid the issue.

MattBrittan (commented):
Thanks for this - I've added a few comments (sorry if these seem fussy!).

paho/pinger.go
Lorderot: Fixed and added a comment. As I understood it, the concern is that it's a potential temporary "memory leak"

MattBrittan: More that it was creating unnecessary activity (sending a `nil` on the channel and then always ignoring the `nil`). I also did suspect that there was a possible route through the code where a go-routine could have been leaked (but did not really follow through investigating this as it's simpler just to remove the unnecessary send).

PR #262, opened 2024-08-11
Add onPublishComplete callback

shaileshahuja:
Basic implementation for discussion

### DIFF
--- a/paho/client.go
+++ b/paho/client.go
@@ -833,7 +833,8 @@ const (
 // PublishOptions enables the behaviour of Publish to be modified
 type PublishOptions struct {
 	// Method enables a degree of control over how  PublishWithOptions operates
-	Method PublishMethod
+	Method     PublishMethod
+	OnComplete func(*PublishResponse, error)
 }
 
 // PublishWithOptions is used to send a publication to the MQTT server (with options to customise its behaviour)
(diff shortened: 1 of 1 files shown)


paho/client.go
```diff
@@ -897,15 +898,29 @@ func (c *Client) publishQoS12(ctx context.Context, pb *packets.Publish, o Publis
 	if _, err := pb.WriteTo(c.config.Conn); err != nil {
 		c.debug.Printf("failed to write packet %d to connection: %s", pb.PacketID, err)
 		if o.Method == PublishMethod_AsyncSend {
-			return nil, ErrNetworkErrorAfterStored // Async send, so we don't wait for the response (may add callbacks in the future to enable user to obtain status)
+			if o.OnComplete != nil {
+				go o.OnComplete(nil, fmt.Errorf("failed to write packet to connection: %w", err))
```
MattBrittan, line 902: The packet will be sent when the connection is reestablished (assuming the session is not cleared). That's really the point of returning this error - it means "I could not send the packet right now but it's in the session state". If the callback will be called here then I think the same error should be passed (or it should wrap `ErrNetworkErrorAfterStored`).

paho/client.go
```diff
@@ -897,15 +898,29 @@ func (c *Client) publishQoS12(ctx context.Context, pb *packets.Publish, o Publis
...
+		return nil, fmt.Errorf("failed to write packet to connection: %w", err)
 	}
 	c.config.PingHandler.PacketSent()
 
 	if o.Method == PublishMethod_AsyncSend {
-		return nil, nil // Async send, so we don't wait for the response (may add callbacks in the future to enable user to obtain status)
+		if o.OnComplete != nil {
+			go func() {
+				pubResp, err := c.waitForResponse(pubCtx, pb, ret)
+				o.OnComplete(pubResp, err)
```
MattBrittan, line 914: This works but will not survive a reconnection so I'm unsure if it offers much benefit. Will add general comments.

MattBrittan (commented):
A user could replicate this using the current @master by calling the non-async version of `PublishWithOptions` and calling the callback themselves when done. So my feeling is that adding this option increases the library complexity, but does not really allow the user to do anything new.

We may have been talking at cross purposes in the discussion area. I see this functionality as being much more useful (and tricky to implement!) in Autopaho because the connection may drop between the message being sent and the process completed. `PublishMethod` was added specifically to allow `autopaho` to implement `PublishViaQueue` (because it needs to know if/when the message has been moved into the session state, meaning it can be removed from the queue).

When adding anything to `paho` I think to myself "how will I use this in `autopaho`"; there are a few reasons for this:
* Most users will use autopaho (because they don't want to write connection management code themselves)
* If they don't use autopaho then they will, most likely, face similar challanges (i.e. how will this work if a reconnect is needed).

I cannot currently see how this PR would be applied in autopaho; i.e. how can I call `PublishViaQueue`/`autopaho.Publish` and receive a notification when the message has been delivered (or we know it will not be delivered). Implementing that will be considerably more work.

shaileshahuja:
That makes sense, will close this PR and think about how to implement this functionality in autopaho.

PR #285, opened 2024-12-11
feat: add context support to router

johndeng:
- Add context parameter to Router.Route method
- Add context parameter to router handlers
- Update documentation and examples
- Add context propagation tests

This change allows proper context handling in message routing, enabling timeout control and value propagation through the message handling chain.

### DIFF
--- a/autopaho/examples/router/router.go
+++ b/autopaho/examples/router/router.go
@@ -44,7 +44,9 @@ func main() {
 	}
 
 	router := paho.NewStandardRouter()
-	router.DefaultHandler(func(p *paho.Publish) { fmt.Printf("defaulthandler received message with topic: %s\n", p.Topic) })
+	router.DefaultHandler(func(c context.Context, p *paho.Publish) {
+		fmt.Printf("defaulthandler received message with topic: %s\n", p.Topic)
+	})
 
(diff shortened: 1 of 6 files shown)


MattBrittan:
Could you please raise an issue saying what issue you are trying to solve with this PR.

At first glance I'm not sure that this is a good idea for a couple of reasons (after a glance at the code - it's difficult to say much as I don't really know what problem you are trying to solve with this):
 * `context.TODO()` is a signal to the code author that action is needed, it has no real place in a library like this (it's fine while you are developing but I'd expect these to be removed before a PR is raised).
 * It would be expected that the context would be provided by the library, but it's unclear what should be used currently (you may plan to add more but it would be better to discuss this first)
 * generally handlers should execute rapidly and, in most cases, should not stop when the client is disconnecting (because this would lead to loss of messages). So if this change is to happen it will need clear documentation.

johndeng:
Sorry for any confusion this PR may have caused. 

Let's move this discussion to #286 , I'll close this PR for now.

MattBrittan:
Thanks - and no worries. I'll add some comments on the issue you raised, I'm open to changes to the library but as it's difficult to remove a feature once it's added some care is needed.

PR #298, opened 2025-04-24
use ctx.Err() instead of a new "context ended" error

nvx:
when the context ends `autopaho` rpc handler returns `fmt.Errorf("context ended")` instead of the more usually expected `ctx.Err()` value.

the corresponding implementation in `paho` already correctly returns `ctx.Err()` in such cases

### DIFF
--- a/autopaho/extensions/rpc/rpc.go
+++ b/autopaho/extensions/rpc/rpc.go
@@ -102,7 +102,7 @@ func (h *Handler) Request(ctx context.Context, pb *paho.Publish) (resp *paho.Pub
 	select {
 	case <-ctx.Done():
 		_ = h.getCorrelIDChan(string(pb.Properties.CorrelationData))
-		return nil, fmt.Errorf("context ended")
+		return nil, ctx.Err()
 	case resp = <-rChan:
 		return resp, nil
 	}


nvx:
I tried hitting the "Sign" button under ECA at https://api.eclipse.org/git/eca/status/gh/eclipse-paho/paho.golang/298 which linked to https://accounts.eclipse.org/user/eca but it's throwing an `Application is not available` error currently

MattBrittan:
> I tried hitting the "Sign" button under ECA at https://api.eclipse.org/git/eca/status/gh/eclipse-paho/paho.golang/298 which linked to https://accounts.eclipse.org/user/eca but it's throwing an `Application is not available` error currently

Unfortunately there is an ongoing outage at eclipse, they are working to resolve this and I'd expect services to be up by Monday so please retry then (status at https://www.eclipsestatus.io/). Unfortunately I cannot accept the PR without this (think the change makes sense).

nvx:
It looks like the  login page is now working, but the only sign up link I can find sends me to the homepage and the login page doesn't appear to have a sign up button. Are signups still broken or?

MattBrittan:
> It looks like the login page is now working, but the only sign up link I can find sends me to the homepage and the login page doesn't appear to have a sign up button. Are signups still broken or?

Last info I have is:

> While investigating a performance issue with our high-availability storage cluster, that same technology which was intended to not fail has, disappointingly, failed. The team is working feverishly to restore the cluster, and in the interim, is reverting to legacy storage to restore some services more rapidly.

> Most of our services have been restored to legacy storage. However, the state of the storage cluster is not good, and we will likely need to abandon it. Starting Monday in CET, the EF IT team will enact Plan B: to provision enough legacy storage and begin restoring Nexus ([repo.eclipse.org](http://repo.eclipse.org/)) and each of the 250+ Jenkins instances so that your work can resume.

> As the current clustered filesystem is inaccessible, we will need to restore some content from virtual tape. There may be some data loss related to build jobs. We will continue to monitor Plan A (the storage cluster recovery) throughout the weekend, but we are not optimistic.

> Once service is restored, we will assemble and share a comprehensive postmortem.

I suspect it might be a few days, the ECA is really the only service this repo relies upon so the only real impact here is what you are seeing. Please try again later in the week (I probably would not get the PR merged until the end of the week anyway).

MattBrittan:
FYI - looks like the ECA system is back up and running again.

PR #299, opened 2025-05-02
add constants for `RetainHandling` subscribe option

agebhar1:
Improve code readability by avoid 'Magic Numbers'.


MattBrittan:
Thanks - makes sense to use constantss here. I'll accept as-is but would really prefer that the constants be explicit (rather than using `iota`) for consistancy with other files and to make the values obvious when reading the code.

PR #307, opened 2025-07-17
Remove extra cla check

st3v3nmw:
Seems we now have [a CLA check workflow](https://github.com/canonical/cla-workflow-hosting/blob/main/.github/workflows/cla.yml) enabled org-wide so we no longer need this.


st3v3nmw:
oops 🤦🏾

PR #309, opened 2025-07-20
fix folder existence check in autopaho file queue

### LINKED ISSUE #308: Autopaho file queue folder existance check is wrong
The autopaho file queue is designed to create the target folder if missing. Two lines are switch so no new folder will ever be created and return the error `stat on folder failed`.

The lines in [queue.go](https://github.com/eclipse-paho/paho.golang/blob/master/autopaho/queue/file/queue.go) must be switched:

The original code:

```go
folderExists := true
if _, err := os.Stat(path); err != nil {
	if os.IsNotExist(err) {
		folderExists = false
	}
	return nil, fmt.Errorf("stat on folder failed: %w", err)
}
```

New code:

```go
folderExists := true
if _, err := os.Stat(path); err != nil {
	if !os.IsNotExist(err) {
		return nil, fmt.Errorf("stat on folder failed: %w", err)
	}
	folderExists = false
}
```

**To reproduce**
```go
package main

import (
	"log"

	"github.com/eclipse/paho.golang/paho/store/file"
)

func main() {
	_, err := file.New("does-not-exist", "prefix", "ext")
	if err != nil {
		log.Fatalf("failed: %v\n", err)
	}
}
```

**Debug output**
```
failed: stat on folder failed: stat does-not-exist: no such file or directory
exit status 1
```

**Expected behaviour**
Created missing folder `does-not-exist`.

**Software used:**
- Client version: 0.22

**Additional context**

alpine-ibex:
Fix #308 to do the intended behaviour: create missing folders in autopaho file queue.

**Closing issues**

closes #308

### DIFF
--- a/autopaho/queue/file/queue.go
+++ b/autopaho/queue/file/queue.go
@@ -64,8 +64,9 @@ func New(path string, prefix string, extension string) (*Queue, error) {
 	if _, err := os.Stat(path); err != nil {
 		if os.IsNotExist(err) {
 			folderExists = false
+		} else {
+			return nil, fmt.Errorf("stat on folder failed: %w", err)
 		}
-		return nil, fmt.Errorf("stat on folder failed: %w", err)
 	}
 	if !folderExists {
 		if err := os.MkdirAll(path, folderPermissions); err != nil {


MattBrittan:
Thanks for the PR.

PR #310, opened 2025-07-21
drop redundant stat before MkdirAll

alpine-ibex:
`MkdirAll` does nothing and returns nil if path exists. So there is no need to check if the path exists.


MattBrittan:
Love removing code!

PR #315, opened 2025-09-05
Fix AuthResponse success value being set to true on error

### LINKED ISSUE #314: Authenticate returns success even on failure
**Describe the bug**
If we send an AUTH packet using paho.Client.Authenticate and the server sends a DISCONNECT in response indicating a failure, the returned AuthResponse struct still sets the Success field to true, when it should be false.

**To reproduce**
Connect to any broker, call `cm.Authenticate` using any invalid credentials, and inspect the returned AuthResponse. The Success field is set to true.

**Expected behaviour**
The Success field of the AuthResponse should be false iff the server rejects our AUTH attempt by sending us a DISCONNECT packet.

**Software used:**
- Client version: v0.22.0

nickajacks1:
The Success field is set to true even when the server denies your AUTH packet. Simply set this to false when a DISCONNECT response is received.

**Testing**

Added unit test that covers the scenario.

**Closing issues**

closes #314

### DIFF
--- a/paho/client_test.go
+++ b/paho/client_test.go
@@ -677,11 +677,24 @@ func TestReceiveServerDisconnect(t *testing.T) {
 }
 
 func TestAuthenticate(t *testing.T) {
-	clientLogger := paholog.NewTestLogger(t, "Authenticate:")
-	ts := basictestserver.New(paholog.NewTestLogger(t, "TestServer:"))
-	ts.SetResponse(packets.AUTH, &packets.Auth{
-		ReasonCode: packets.AuthSuccess,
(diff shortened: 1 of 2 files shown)


MattBrittan:
Makes sense, thanks for the PR. This is functionality that I don't use, but based on a quick look I feel it could probably do with some work (for example, what happens in `Authenticate` if the connection drops, it appears to rely on the `ctx` passed in which is not ideal). Improved docs would also be good (returning a `AuthResponse` when a `DISCONNECT` has been received may not be the expected outcome). Would welcome any further improvements.

Issue #71, opened 2021-07-09
routing based on Subscription Identifier


samba:
From reading the MQTTv5 specs and other implementations, it seems the `Subscription Identifier` is [intended to facilitate routing][1], when a client has multiple subscriptions that would match a specific message's topic. The broker forwards the message with the attached `Subscription Identifier`, which is provided originally by the client, and should enable the client to associate each forwarded message with the accompanying subscription handler. 

https://github.com/eclipse/paho.golang/blob/a215349036a5852a292890e4fea23585aa19e747/paho/router.go#L93-L97

Seems like this would be the right place to implement that Subscription Identifier logic, to select which routes are applicable for a given message. 

Are there historical reasons why this approach has not been implemented in this library?
Would there be any objection to implementing that approach in the near future?

[1]: https://www.emqx.io/blog/subscription-identifier-and-subscription-options

alsm:
No, there is no particular reason this hasn't been implemented and I am more than happy to have a PR with this function, or look at it myself.

zyfdegh:
Got same problem, a **new tag** need to be released please @alsm 

I'm using v0.10.0 but the router has issue matching shared subscription topic, and it was fixed in this commit
CommitID: fd33e6d2ba8489a1fe9a3d36b447c2198d0af427

Currently switching to master branch solves.

MattBrittan:
Quick summary as the link in the original question is dead...

With MQTT when subscribing you can pass a [subscription identifier](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901166) (an integer between 1 and 268,435,455). When the broker sends a PUBLISH packet it will include the subscription identifier (for the SUBSCRIBE that led to the PUBLISH).

This seems like a fairly niche requirement, which would add complexity (i.e. would need a separate  `make(map[subIdType][]MessageHandler),` and way for the IDs to be registered). As things stand it is fairly easy for users to implement this themselves (take a copy of `paho.StandardRouter` and modify as required; the subscription identifier is available within the `*packets.Publish` passed to `Route`), so my feeling is that there would be little value in adding this to the library.

MattBrittan:
I'm going to close this due to lack of activity. It's a fairly niche requirement and can be implemented via a custom router with minimal effort.

PR #321, opened 2025-11-23
Add Context Router with Middleware Support

### LINKED ISSUE #286: Add context.Context support to StandardRouter message routing
Currently using paho.StandardRouter for message routing but it lacks context.Context support. 

Adding context support would allow applications to integrate distributed tracing (OpenTelemetry) much easier.

akfaiz:
### Description

This PR introduces a context-aware router extension (`paho/extensions/routercontext/`) that provides topic-based message routing with middleware support. Handlers receive `context.Context` for cancellation and timeout control.

⚠️ **Note**: This is an experimental extension, not part of core. Not covered by semantic versioning and support is limited.

### Key Features

- **Context-Aware Handlers**: Handlers receive `context.Context` along with the message
- **Middleware Support**: Wrap handlers for logging, metrics, error handling, etc.
- **MQTT Wildcards**: Full support for `+` (single-level) and `#` (multi-level) patterns
- **Thread-Safe**: Uses `sync.RWMutex` for concurrent access
- **Default Handler**: Route unmatched topics to a default handler

### Changes

**Extension** (`paho/extensions/routercontext/`):
- `Router` struct with handler registration and routing
- `Handler` and `Middleware` types
- Topic matching with MQTT wildcard support
- Default handler and debug logging support

**Examples** (`autopaho/examples/routercontext/`):
- Complete working example with autopaho integration
- Middleware implementations: logger and panic recoverer
- Comprehensive README with usage guide

**Tests** (`paho/extensions/routercontext/routercontext_test.go`):
- 23 unit tests covering all functionality
- 97.1% code coverage
- Tests for handler routing, middleware ordering, wildcards, and concurrency

**Documentation**:
- Inline code comments in all files
- README in extension with API reference and best practices
- README in examples with integration guide
- Stability warnings on both READMEs

### Usage

```go
router := routercontext.NewRouter()
router.Use(middleware.Logger, middleware.Recoverer)

router.RegisterHandler("sensors/#", func(ctx context.Context, p *paho.Publish) {
    slog.InfoContext(ctx, "Message received", "topic", p.Topic)
})

router.DefaultHandler(func(ctx context.Context, p *paho.Publish) {
    slog.InfoContext(ctx, "Unmatched", "topic", p.Topic)
})
```

### Testing

- ✅ All 23 tests passing
- ✅ 97.1% code coverage
- ✅ All existing tests pass without modification

### Closes

closes #286

**ECA Compliance**: Commits are signed with an ECA'd email address.


paho/client.go
```diff
@@ -75,6 +75,11 @@ type (
 		// slice (which provides a more flexible approach to handling incoming messages).
 		Router Router
 
+		// ContextRouter is an alternative message router that receives context.Context along with the message.
+		// It is used when Router is not set, allowing context-aware message handling for operations
+		// that require context propagation, cancellation, or timeout control.
+		ContextRouter ContextRouter
```
MattBrittan: `Router Router` notes that it's depreciated so I'm not keen to see an additional router added here (the idea is that users utilise `OnPublishReceived` (the README notes "ClientConfig.Router will be removed in a future release"). There are a few reasons for this change, but the main one is that its difficult for us (library authors) to predict how this will be used (and example being routing based on Subscription identifier - see #168 for some discussion on this).

MattBrittan (changes requested):
I've added a few comments to your code. My main point here is that I don't think this should require changes in `client` (because `client.Router` is depreciated so you should not try to duplicate this). 

Having reviewed this I'm still a bit uncertain about it's utility - it adds quite a bit of code that could be fairly quickly handled in `OnPublishReceived` (without the need for the added overhead). My initial thought is that this library should only provide a basic router that will meet the majority of needs (to help users get started and provide an example) - users with more complex requirements can use `OnPublishReceived` to do whatever they need.

Note that my reluctance here is not really related to your code. Experience from the V3 client indicates that we should try to keep the client as small as possible (we need to support anything added so want to limit additions to stuff that most users need, whist providing ways for other users to meet their unique needs themselves). The reason for this approach is that if I accept this PR the next request may well be to add support for something like #71 (which is a valid request but not common enough to be worth adding here - we cannot be all things to all users).

I will be very interested in feedback from others reading this PR because othere may see this as an essential addition (it's not something that I would use myself).

Currently my feeling is that it might be more appropriate to:
  * Add a `readme` to `paho/extensions`  & `autopaho/extensions`  stating that items in this folder are not part of the core library, are not covered by semantic versioning, may be broken by updates eleswhere, support will be limited etc.
  * Move this PR into those folder (after removing the changes to `client.go` etc).
  * Possibly reference it in the main README (I do need to update that anyway)

nickajacks1:
@MattBrittan here's my 2 cents.

The StandardRouter type itself is useful enough to where I basically use it as the only handler right now. The topic matching is useful. That said, I haven't tried to (or needed to (yet)) extend the StandardRouter for further purposes. I think your instincts are right that people's use cases will be varied enough that it's probably wiser to pursue extensibility over thorough functionality. To that end, it might be better to focus on having composable building blocks. One way that would make it easier for people to implement their own router-like constructs would be to make the route matching logic exported. The StandardRouter doesn't actually do a whole lot more than loop through the registered handlers, check if they match, then call all the handlers for said subscription.  By having access to the match function, it would be pretty easy to implement your own router (such as one that passes along a context).

One microcosm of the varied use cases is that the fact that this router uses context.Background means that I wouldn't be able to use it effectively for my case. My applications always pass in a single parent context from the main function for cancellation purposes. This implementation does not provide that, but maybe that's perfectly fine for the author's case. Even though I need to pass context along to a lot of my handlers, I wouldn't be able to use this router to accomplish it. For middleware, it would be perfectly fine for me to simply add additional handlers to the front of the `OnPublishReceived` chain, but for this author, that might not be good enough. 

Just brainstorming on my own a bit here, but a good way to nudge users into the right direction might be to build out a combination of documentation and examples to show various ways to accomplish different cases. I think mosquitto does this well for its plugin API. There are a lot of examples that show what you can accomplish, and it's a good justification to avoid adding too many bells and whistles to the API.

As far as accepting this Router, I would say that's really up to you on how you want to handle extensions. If this were my project, I might request that this be moved to be included only as an example, or be added to some contrib repo.
As a user of this project, what I really wouldn't like is if extensions started to add extra dependencies to go.mod that I don't need. That's not the case here, but as you allude, it's something of a slippery slope.

akfaiz:
I agree that using `context.Background()` inside `Route` is not ideal, since it prevents applications from passing their own parent context (for example, one derived from OS signals).
To improve this, we can provide a new `RouteContext` method that accepts a context, and let the existing `Route` method simply fall back to `context.Background()` for backward compatibility.

Proposed API:
```
// Backward-compatible entrypoint
func (r *Router) Route(pb *packets.Publish) {
    r.RouteContext(context.Background(), pb)
}

// New context-aware version
func (r *Router) RouteContext(ctx context.Context, pb *packets.Publish) {
    // routing logic using the provided ctx
}
```

This allows applications to pass their own parent context (e.g., from signal.NotifyContext), while keeping the existing Route signature working as before.

MattBrittan:
> // Backward-compatible entrypoint

I would not really worry about this. The `Router` interface really dates back to the legacy `Router` in `ClientConfig`. With the move to `OnPublishReceived` I don't see much benefit in adhering to this (and suspect that most end users would want to push in their own `context`).

Other that that I think that this change would make your PR easier to use/understand. I think quite a few devs think of `Context` as mostly being mostly about cancellation (I very rarely use it to pass values) so having the router create the context did feel a bit wrong. Making it user created also avoids confusion about it including cancellation signals from the library (something mentioned in issue #286, but I was never clear what context would be passed in that case, so think this approach is better).

paho/extensions/routercontext/routercontext.go
```diff
@@ -0,0 +1,211 @@
...
+		ctx = context.Background()
+	}
+
+	var topic string
+	if pb.Properties.TopicAlias != nil {
+		r.debug.Println("message is using topic aliasing")
+		if pb.Topic != "" {
+			// Register new alias
+			r.debug.Printf("registering new topic alias '%d' for topic '%s'", *pb.Properties.TopicAlias, m.Topic)
+			r.aliases[*pb.Properties.TopicAlias] = pb.Topic
```
MattBrittan, line 106: Bug (duplicated from main router) - we only have an RLock so should not be modifying aliases. Will raise a separate issue for this (noting it as this was picked up whilst reviewing this PR)

paho/extensions/routercontext/routercontext_test.go
```diff
@@ -0,0 +1,474 @@
...
+	}
+}
+
+func Test_ConcurrentRouting(t *testing.T) {
+	r := NewRouter()
+	var callCount int
+	done := make(chan bool)
+
+	handler := func(ctx context.Context, p *paho.Publish) {
+		callCount++
```
MattBrittan, line 429: Data race? callCount will be incremented from multiple goroutines (`Route` only holds a read lock so believe there is a race).

paho/extensions/routercontext/routercontext.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+		if t, ok := r.aliases[*pb.Properties.TopicAlias]; ok {
+			r.debug.Printf("aliased topic '%d' translates to '%s'", *pb.Properties.TopicAlias, m.Topic)
+			topic = t
```
MattBrittan, line 110: This code (from original) seems likely to cause confusion because it resolves the alias but the message passed to the handler (`m`) will still have the alias. I'll log a separate issue re this as its an issue in the existing router too.

MattBrittan (changes requested):
Really just the race condition (others are copied from the current router).

MattBrittan:
Apologies for the massive delay on this. I think including this as an example makes sense. I think there is a data race in the test so if you can resolve that I'll accept the PR.

PR #323, opened 2025-12-01
Check for invalid Variable Byte Integer when encoding/decoding

MattBrittan:
The Variable Byte Integer header can only be four bytes long; this limits the maximum length of the encoded data to 268,435,455 bytes. We should check this when decoding (important) and encoding (not so important but not checking could result in an invalid packet being sent). Checking for this required further changes as the library previously assumed that there are no errs on encoding (previously would panic if VBI was over length).


MattBrittan:
Closes #324 

Issue raised after PR due to discussions re potential security implications.

PR #329, opened 2026-05-24
fix(rpc): free correlation entry when a request gets no response

### LINKED ISSUE #313: Potential memory leak in RPC request
**Describe the bug**
The RPC Request() can return after an error publishing a message, but the channel map, h.correlData, will still contain the channel keyed by the correlation ID. The published message may or may not get through to the broker and the recipient.

- If it does, then the channel map entry is cleared in responseHandler(), and the response is handled by bug fix https://github.com/eclipse-paho/paho.golang/issues/289 so the response does not hang
- If it doesn't, then there will be no response, but the channel map, h.correlData, is never cleared of the channel

**Expected behaviour**
The channel map, h.correlData, should be cleared of the unusable channel before Request(0 returns, maybe in the same way as it is done in the select if the context finishes, with a call
`_ = h.getCorrelIDChan(string(pb.Properties.CorrelationData))`

ultramcu:
## Summary

Closes #313.

`rpc.Handler.Request` registers a correlation entry (`cID -> response channel`) in `h.correlData`, but that entry is only ever removed by `responseHandler` — i.e. when a matching response actually arrives. Any request that:

- times out / has its context cancelled (`<-ctx.Done()`), or
- fails to publish (`h.c.Publish` returns an error), or
- never receives a response

leaves its entry in `h.correlData` forever. Over a long-running session with any lost/slow responses this is a steady memory leak (one map entry + channel per affected request).

The earlier buffered-channel change fixed the goroutine leak on cancellation, but the map entry itself was still leaked.

## Fix

Add `removeCorrelID` and `defer` it at the top of `Request`, so the correlation entry is removed on every exit path (response received, timeout, cancel, publish error). `delete` on an already-removed key is a no-op, so this is safe alongside `responseHandler`'s removal.

## Testing

Added `TestRequestCleansUpCorrelDataOnTimeout` (uses the existing `basictestserver`): connect, create the handler, then issue a `Request` that is acked at the PUBLISH level but never gets a response on the response topic, so it times out. It asserts `len(h.correlData) == 0` afterwards.

- Verified the test **fails without** this change (`correlData` length is 1) and **passes with** it.
- `go test ./paho/extensions/rpc/ -race` passes; `go vet` clean; `gofmt` clean.

## Note (separate, not addressed here)

`Request` derives the correlation ID from `time.Now().UnixNano()`, so two concurrent requests issued within the same nanosecond can collide on the same `cID` (one overwrites the other's channel, and that caller then hangs until its context fires). Happy to follow up with a counter/UUID-based ID in a separate PR if you'd like — kept this one focused on the #313 leak.

### DIFF
--- a/paho/extensions/rpc/rpc.go
+++ b/paho/extensions/rpc/rpc.go
@@ -76,11 +76,22 @@ func (h *Handler) getCorrelIDChan(cID string) chan *paho.Publish {
 	return rChan
 }
 
+func (h *Handler) removeCorrelID(cID string) {
+	h.Lock()
+	defer h.Unlock()
+
+	delete(h.correlData, cID)
+}
+
 func (h *Handler) Request(ctx context.Context, pb *paho.Publish) (*paho.Publish, error) {
 	cID := fmt.Sprintf("%d", time.Now().UnixNano())
 	rChan := make(chan *paho.Publish, 1) // Buffered to prevent goroutine leak when context cancelled
 
 	h.addCorrelID(cID, rChan)
+	// Ensure the correlation entry is always removed. responseHandler only
+	// removes it when a matching response arrives, so without this a request
+	// that times out, is cancelled, or fails to publish leaks a map entry.
+	defer h.removeCorrelID(cID)
 
 	if pb.Properties == nil {
 		pb.Properties = &paho.PublishProperties{}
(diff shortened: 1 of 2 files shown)


MattBrittan:
Looks good to me.  Can see your point re the correlation ID. It may make more sense to generate the ID in `addCorrelID` because that way it can check for uniqueness (has the lock so free access to the map). Using the time as a starting point makes some sense, using a counter could lead to collisions following re-connection (messages queued on the broker etc).

PR #330, opened 2026-05-24
fix(state): use a signed send quota so retransmit cannot underflow

ultramcu:
## Summary

`sendQuota.quota` and `initialQuota` are `uint16`, but the send-quota was designed around a counter that can go **negative** — the code's own comments say so (`// Note: can go < 0 if noWait used` in `acquire`, and `// Possible quota could go negative when using noWait` in `Release`). On an unsigned type it can't go negative; it wraps.

`Retransmit()` calls `acquire(ctx, noWait=true)`, which unconditionally does `s.quota--` without blocking. On a reconnect, `ConAckReceived` creates a fresh `newSendQuota(recvMax)` and then calls `Retransmit()` once per stored (unacknowledged) message (`state.go`). If there are more stored messages than the receive maximum, the quota is decremented past zero and **wraps to ~65535**. After that:

- `Acquire` always sees `quota > 0`, so it stops blocking — the broker's **Receive Maximum is no longer honoured** (the client can exceed it).
- `Release` is broken: `s.quota < s.initialQuota` is false (65535 < recvMax), so it never decrements back and returns `ErrUnexpectedRelease` indefinitely; the quota never recovers.

### Trigger (normal usage)

A client that buffers several QoS 1/2 publishes while disconnected and then reconnects, where the backlog exceeds the broker's (possibly small) receive maximum.

## Fix

Make `quota` and `initialQuota` signed (`int`) so the counter can legitimately go negative (over-allocated) as the existing comments intend, instead of wrapping. The existing `Acquire`/`Release` logic (`quota >= 0`, `quota < initialQuota`) then behaves correctly: releases bring the over-allocated quota back up to the initial value.

## Testing

Added `TestQuotaRetransmitDoesNotUnderflow`: `newSendQuota(2)`, `Retransmit()` 3 times (one past the max), then `Release()` must succeed back to the initial value (rather than returning `ErrUnexpectedRelease`). Verified it **fails before** this change and **passes after**. The existing quota tests still pass.

- `go test ./paho/session/state/ -race` and `go test ./paho/ -race` pass; `gofmt` clean.

### DIFF
--- a/paho/session/state/sendquota.go
+++ b/paho/session/state/sendquota.go
@@ -46,16 +46,21 @@ var ErrUnexpectedRelease = errors.New("release called when quota at initial valu
 
 // newSendQuota creates a new tracker limited to quota concurrent messages
 func newSendQuota(quota uint16) *sendQuota {
-	w := &sendQuota{initialQuota: quota, quota: quota}
+	w := &sendQuota{initialQuota: int(quota), quota: int(quota)}
 	return w
 }
 
 // sendQuota provides a way to bound concurrent access to a resource.
 // The callers can request access with a given weight.
+//
+// quota is a signed counter: Retransmit (noWait) decrements it without blocking, so on a
+// reconnect that retransmits more stored messages than the receive maximum it can legitimately
+// go negative (over-allocated). It must therefore be a signed type — a uint16 here would wrap to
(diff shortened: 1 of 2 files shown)


MattBrittan:
Good catch - thanks for finding/fixing this!

PR #334, opened 2026-05-30
fix(state): remove stray debug fmt.Println from sendQuota Acquire

### LINKED ISSUE #333: Stray debug fmt.Println in sendQuota.Acquire writes to stdout
### Describe the bug

`paho/session/state/sendquota.go`, inside `sendQuota.acquire`, the `ctx.Done()` branch's inner `case <-ready:` arm contains a stray debug print:

```go
case <-ready:
    // Acquired the semaphore after we were cancelled. Rather than trying to
    // fix up the queue, just pretend we didn't notice the cancellation.
    err = nil
    fmt.Println("quota released entry but ready so nil error ", s.quota)
```

This arm is reached when a waiter's context is cancelled but a slot is `Release`d at almost the same moment, so the semaphore becomes ready right after the cancellation. When that happens, the library writes debug output directly to the consuming application's **stdout**.

### Impact

Any program using the v5 `paho` client can get unexpected
`quota released entry but ready so nil error  <n>` lines printed to **stdout**
under normal operation — specifically when an `Acquire` whose context is
cancelled races with a `Release` (which is a legitimate, reachable scenario). A
library should not write to stdout; this pollutes the consumer's output.

### To reproduce

Exhaust the send quota so a caller becomes a waiter, then cancel that caller's
context at the same moment another goroutine releases a slot. Driving this race
in a loop reliably hits the branch and prints to stdout.

### Proposed fix

Remove the `fmt.Println` line (and the now-unused `"fmt"` import). Behaviour is
unchanged — `err = nil` and all control flow stay the same. I'd also add a
regression test that captures stdout while driving the cancel/ready race and
asserts the library writes nothing (fail-before/pass-after verified locally).

Happy to open a PR.

ultramcu:
Closes #333.

The `case <-ready:` arm of the `ctx.Done()` select in `sendQuota.acquire`
(`paho/session/state/sendquota.go`) — reached when a waiter's context is
cancelled but a slot is `Release`d at almost the same moment, so the semaphore
becomes ready right after the cancellation — contained a stray `fmt.Println(...)`
that wrote debug output to the consumer's **stdout** whenever that race was hit.
A library should not print to stdout.

This removes that line and the now-unused `"fmt"` import. Behaviour is
unchanged — `err = nil` and all control flow stay the same.

**Testing**

Adds `TestSendQuotaAcquireSilentOnCancelReadyRace` (in-package), which captures
`os.Stdout` while driving the cancel-vs-`Release` race and asserts the library
writes nothing:

- Fail-before: re-inserting the `Println` makes the test capture thousands of
  bytes of stray stdout and fail.
- Pass-after: passes, race-clean under `-race`, and non-flaky across many runs.

`make test` passes.

### DIFF
--- a/paho/session/state/sendquota.go
+++ b/paho/session/state/sendquota.go
@@ -18,7 +18,6 @@ package state
 import (
 	"context"
 	"errors"
-	"fmt"
 	"sync"
 )
 
@@ -102,7 +101,6 @@ func (s *sendQuota) acquire(ctx context.Context, noWait bool) error {
 			// Acquired the semaphore after we were cancelled. Rather than trying to
 			// fix up the queue, just pretend we didn't notice the cancellation.
 			err = nil
-			fmt.Println("quota released entry but ready so nil error ", s.quota)
 		default:
 			// Remove ourselves from the list of waiters
 			for i, r := range s.waiters {
--- a/paho/session/state/sendquota_strayprint_test.go
+++ b/paho/session/state/sendquota_strayprint_test.go
@@ -1,0 +1,118 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ *
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License v2.0
+ *  and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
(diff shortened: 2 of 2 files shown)


MattBrittan:
Sorry for leaving that in!

PR #335, opened 2026-05-30
Seperate topic alias processing from routing

### LINKED ISSUE #331: Router race on `r.aliases` (only if MQTT v5 aliases are used)
`StandardRouter.Route` obtains a read lock (`r.RLock()`) but may then update `r.aliases` (only when MQTT V5 aliases are used). This looks like a race.

### LINKED ISSUE #332: Topic alias - Should router pass msg with alias substituted in message

`StandardRouter` processes topic aliases and uses the resolved alias when determining which handler to call. The message passed to the handler will contain the unresolved topic alias. This may be confusing if the handler uses the topic.

I think the router should probably update the message body (it needs to either handle aliases, or not handle them).

There is not really a right answer here (I've not used topic aliases myself).

**Additional context**

Logging this to see if others have an opinion.

MattBrittan:
This addresses the data race in `paho/router` by separating the handling of aliases. As client now calls each function in `[]func(PublishReceived) (bool, error)`  when a message is received we can add a separate handler before the router that manages aliases.

This approach also means that the message passed to the end user will have the expected `Topic` (currently the user has no way of determining the actual topic if the route pattern contained a wildcard).

Note: Currently a work in progress - feedback welcome!

closes #331 
closes #332


MattBrittan:
Closed this as the PR was too messy...

PR #339, opened 2026-06-01
feat: add ReadPacketN with maximum packet size limit

### LINKED ISSUE #336: paho.ReadPacket Parket size should be limited to `Maximum Packet Size` (if set)
Currently a hostile broker could (potentially) cause a client to run out of memory by sending a lot of 256MB messages. I feel the risk is fairly low because the attacker would need to control the broker, and be able to deliver messages quickly enough to overwhelm the client.

The CONNECT packet can include a [`Maximum Packet Size`](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901050) property. If this has been set then, ideally, `paho.ReadPacket` should reject any packets larger than this (ideally as soon as it receives the packet size so no memory is allocated). As per the spec, upon receiving an oversize packet, "the Client uses DISCONNECT with Reason Code 0x95 (Packet too large), as described in section 4.13".

The cleanest way of implementing this is probably to add a `paho.ReadPacketN` (or similar) that takes the reader and a maximum size. The existing function can then just call the new one (`ReadPacket` is also used when reading from disk).

toller892:
## Summary

Adds `ReadPacketN(r io.Reader, maxSize int)` — a variant of `ReadPacket` that rejects packets exceeding a caller-specified size limit before allocating memory for the payload.

## Motivation (Fixes #336)

A malicious broker could cause a client to run out of memory by sending very large packets (up to 256MB per the MQTT spec). The CONNECT packet's `Maximum Packet Size` property tells the broker the client's limit, but `ReadPacket` never enforced it. This change lets callers pass that limit and get an early rejection.

## Implementation

- `ReadPacketN` reads the fixed header and VBI as before, then checks `1 + vbiLen + remainingLength > maxSize` **before** allocating the payload buffer
- `ReadPacket` now delegates to `ReadPacketN(r, 0)` — full backward compatibility (maxSize=0 means no limit)
- Error message references MQTT 5.0 Reason Code 0x95 (Packet too large) per spec section 4.13

## Tests

- `TestReadPacketNConnect`: no-limit, large-limit, and undersized-limit with a CONNECT packet
- `TestReadPacketNPingreq`: boundary tests (exact match, under limit) with the simplest 2-byte PINGREQ packet

All existing tests pass.

### DIFF
--- a/packets/packets.go
+++ b/packets/packets.go
@@ -241,17 +241,24 @@ func NewControlPacket(t byte) *ControlPacket {
 }
 
 // ReadPacket reads a control packet from a io.Reader and returns a completed
-// struct with the appropriate data
+// struct with the appropriate data.
 func ReadPacket(r io.Reader) (*ControlPacket, error) {
+	return ReadPacketN(r, 0)
+}
+
+// ReadPacketN reads a control packet from an io.Reader and returns a completed
+// struct with the appropriate data. If maxSize is greater than zero and the
+// total packet size (fixed header + variable header + payload) exceeds maxSize,
+// the packet is rejected with an error. This protects against a malicious broker
+// sending oversized packets that could cause the client to run out of memory.
+// As per MQTT v5.0 spec section 4.13, a client should DISCONNECT with Reason
+// Code 0x95 (Packet too large) upon receiving an oversize packet.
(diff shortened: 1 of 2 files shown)


MattBrittan:
Closing this as it looks like #340 is a duplicate.

PR #340, opened 2026-06-01
fix: enforce Maximum Packet Size in ReadPacket

### LINKED ISSUE #336: paho.ReadPacket Parket size should be limited to `Maximum Packet Size` (if set)

toller892:
## Summary

Adds `ReadPacketN` to the `packets` package that accepts a `maxSize uint32` parameter. When `maxSize > 0`, packets exceeding the limit are rejected **before** the payload buffer is allocated, preventing a hostile broker from causing OOM by sending oversized packets.

The client's `incoming()` loop now uses `ReadPacketN` with the server's `MaximumPacketSize` (from CONNACK properties) to enforce the limit.

Fixes #336 (and the long-standing #3).

## Changes

- **`packets/packets.go`**: New `ReadPacketN(r io.Reader, maxSize uint32) (*ControlPacket, error)` function. `ReadPacket` now delegates to `ReadPacketN` with `maxSize=0` (no limit) for backward compatibility.
- **`paho/client.go`**: `incoming()` uses `packets.ReadPacketN(c.config.Conn, c.serverProps.MaximumPacketSize)`.
- **`packets/packets_test.go`**: `TestReadPacketNMaxSize` covering no-limit, above-limit, at-limit, and below-limit cases.

## Notes

- The size check happens after reading the fixed header + VBI but **before** allocating `make([]byte, remainingLength)` — this is the key safety property.
- Per MQTT v5.0 spec §3.2.2.3.10, the server's Maximum Packet Size applies to packets sent from server to client.
- Related: #3, #114 (closed, stale)

### DIFF
--- a/paho/client.go
+++ b/paho/client.go
@@ -489,7 +489,7 @@ func (c *Client) incoming(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		default:
-			recv, err := packets.ReadPacket(c.config.Conn)
+			recv, err := packets.ReadPacketN(c.config.Conn, c.serverProps.MaximumPacketSize)
 			if err != nil {
 				go c.error(err)
 				return
--- a/packets/packets.go
+++ b/packets/packets.go
@@ -241,8 +241,16 @@ func NewControlPacket(t byte) *ControlPacket {
 }
 
 // ReadPacket reads a control packet from a io.Reader and returns a completed
-// struct with the appropriate data
+// struct with the appropriate data.
 func ReadPacket(r io.Reader) (*ControlPacket, error) {
+	return ReadPacketN(r, 0)
+}
+
+// ReadPacketN reads a control packet from an io.Reader and returns a completed
+// struct with the appropriate data. If maxSize is greater than zero, packets
+// exceeding that size are rejected before the payload is allocated (as per
+// MQTT v5.0 spec section 3.2.2.3.10).
+func ReadPacketN(r io.Reader, maxSize uint32) (*ControlPacket, error) {
 	t := [1]byte{}
 	_, err := io.ReadFull(r, t[:])
 	if err != nil {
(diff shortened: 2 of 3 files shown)


paho/client.go
```diff
@@ -489,7 +489,7 @@ func (c *Client) incoming(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		default:
-			recv, err := packets.ReadPacket(c.config.Conn)
+			recv, err := packets.ReadPacketN(c.config.Conn, c.serverProps.MaximumPacketSize)
```
MattBrittan, line 492: I think this should be `c.clientProps.MaximumPacketSize` (set from the outgoing `CONNECT` request - i.e. us telling the server not to send us anything larger). `serverProps` is set from the CONNACK (and represents the largest packet the server will receive).

MattBrittan:
Thanks for this - is there a difference between it and #339? I will need an ECA before I can accept the PR ([see here](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git) for info - yell out if you are having issues.

I think the PR looks mostly OK (initial quick look) - have just flagged one issue.

MattBrittan:
Closing this due to inactivity.

PR #348, opened 2026-09-01
Improve backoff documentation

ViToni:
This PR fixes some typos and attempts for clearer wording regarding the documentation of the backoff implementation.

### DIFF
--- a/autopaho/backoff.go
+++ b/autopaho/backoff.go
@@ -42,23 +42,23 @@ func NewConstantBackoff(delay time.Duration) Backoff {
 // implementation for an exponential backoff
 ////////////////////////////////////////////////////////////////////////////////
 
-// NewExponentialBackoff provides a random duration within a range starting
-// from a fixed min value up to a "moving" max value that increases
-// exponentially for each attempt up to the specified max value.
+// NewExponentialBackoff returns a backoff which provides a random duration
+// within a range starting from a fixed min value up to a "moving" max value
+// that increases exponentially for each attempt up to the specified max value.
 //
 // The "moving" max is computed by multiplying the initial max value with the
 // factor for each attempt up to the specified max value.
 //
 // Configuration parameters:
-//   - minDelay        - lower bound for computed backoff
-//   - maxDelay        - upper bound for computed backoff
-//   - initialMaxDelay - initial max value which will increase exponentially up to the max delay
-//   - factor          - factor for the exponential increase of initial max delay
+//   - minDelay        - absolute lower bound for computed backoff
+//   - maxDelay        - absolute upper bound for computed backoff
+//   - initialMaxDelay - initial upper bound before exponential growth
+//   - factor          - multiplier applied to the upper bound on each attempt
 func NewExponentialBackoff(
-	minDelay time.Duration, // lower bound for computed backoff
-	maxDelay time.Duration, // upper bound for computed backoff
-	initialMaxDelay time.Duration, // initial max value which will increase exponentially up to the max delay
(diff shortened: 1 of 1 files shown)


autopaho/backoff.go
```diff
@@ -42,23 +42,23 @@ func NewConstantBackoff(delay time.Duration) Backoff {
...
+//   - factor          - multiplier applied to the upper bound on each attempt
 func NewExponentialBackoff(
-	minDelay time.Duration, // lower bound for computed backoff
-	maxDelay time.Duration, // upper bound for computed backoff
-	initialMaxDelay time.Duration, // initial max value which wiil incerease exponentially up to the max delay
-	factor float32, // factor for the exponential increase of initial max delay
+	minDelay time.Duration, // absolute lower bound for computed backoff
+	maxDelay time.Duration, // absolute upper bound for computed backoff
+	initialMaxDelay time.Duration, // initial upper bound before exponential growth
+	factor float32, // multiplier applied to the upper bound on each attempt
```
MattBrittan, line 61: I think this is confusing because the factor applies to `initialMaxDelay`, not  `maxDelay` (suspect this was a search/replace?). A preferable comment might be something along the lines of "used to compute each later attempt's maximum delay as min(previous maximum * factor, maxDelay)"

autopaho/backoff.go
```diff
@@ -42,23 +42,23 @@ func NewConstantBackoff(delay time.Duration) Backoff {
...
 //
 // Configuration parameters:
-//   - minDelay        - lower bound for computed backoff
-//   - maxDelay        - upper bound for computed backoff
-//   - initialMaxDelay - initial max value which wiil incerease exponentially up to the max delay
-//   - factor          - factor for the exponential increase of initial max delay
+//   - minDelay        - absolute lower bound for computed backoff
+//   - maxDelay        - absolute upper bound for computed backoff
+//   - initialMaxDelay - initial upper bound before exponential growth
+//   - factor          - multiplier applied to the upper bound on each attempt
```
MattBrittan, line 56: This reads as if the algorithm includes `factor * maxDelay`. Maybe `growth multiplier; for attempt n >= 1, upperBound(n) = min(initialMaxDelay * factor^(n-1), maxDelay)` would be clearer (but may also be too detailed).

autopaho/backoff.go
```diff
@@ -42,23 +42,23 @@ func NewConstantBackoff(delay time.Duration) Backoff {
...
 // The "moving" max is computed by multiplying the initial max value with the
-// factor for each attemt up the specified max value.
+// factor for each attempt up to the specified max value.
 //
 // Configuration parameters:
-//   - minDelay        - lower bound for computed backoff
-//   - maxDelay        - upper bound for computed backoff
-//   - initialMaxDelay - initial max value which wiil incerease exponentially up to the max delay
-//   - factor          - factor for the exponential increase of initial max delay
+//   - minDelay        - absolute lower bound for computed backoff
```
MattBrittan, line 53: Not sure that adding absolute here really helps (technically it's rounded to ms so is not the absolute min). Personally I feel that `backoff` is confusing here - maybe "minimum delay between retry attempts, at millisecond precision" would be easier to understand?

MattBrittan (commented):
Had been going to approve this but am concerned that "multiplier applied to the upper bound on each attempt" makes it more confusing (because you have an "absolute upper bound" and a "initial upper bound"). 
As such I've held off and made some suggestions (which you should feel free to ignore, but I think the comments at the start of `NewExponentialBackoff` do need some work as I feel they are less clear than what was there previously).

